PR #33, opened 2018-03-20
Fix sample yaml indentation

shakuzen:


eaglerainbow:
Thanks. With all that documentation stuff around I missed that.

PR #46, opened 2018-04-25
Resolves #35 dynamic target list

### LINKED ISSUE #35: Dynamic target list / Regex in target configuration
I haven’t tried using promregator quite yet but it looks very promising and is very well documented- good work to the people behind it.

In my use case, I’d like to be able to have promregator scrape from all apps running on CF without having to manually update the config and restart promregator every time there’s a new app on CF. I don’t believe this use case is described in the documentation.

Does promregator reload the config periodically. If so, one way this can be accomplished is to have anindependent side process doing service discovery on what appnames exist across all orgs and space, and then write them to the promregator config file- similar to how Prometheus handles its file-based service discovery.

Another solution may be to have the promregator config support regex for the targets. That way I can match against all orgs/spaces/apps. This would force Promregator to periodically check all existing orgs/spaces/apps, and so this may or may not impact performance somewhat.

If there’s interest in having a built in solution to this, like using regex, and if there’s an idea in how it should be implemented, I’d consider PRing it.

tzolov:
Resolves #35
Pragmatic extension to allow dynamic targets detection.  

It relies on a name convention to signal `promregator` when to "expand" a single target definition (with applicationName set to `SPACE_ALL_APPLICATIONS`) into multiple targets definitions - one for every  app in that org/space. 

For example the following target configuration:
```
  targets:
    - orgName: scdf-ci
      spaceName: space-christian
      path: '/actuator/prometheus' # Spring Boot micrometer actuator entry point
      applicationName: 'SPACE_ALL_APPLICATIONS'
```

Will dynamically expand the targets definitions to all applications running in the scdf-ci/space-crhistian space. 

In this PR I've extended the [ReactiveCFAccessorImpl#retrieveApplicationId](https://github.com/promregator/promregator/compare/master...tzolov:gh-35-dynamic-target-list?expand=1#diff-cd96d2c1b0fda9ca65f7a8920bd54f90R162) to drop the name filtering in case of SPACE_ALL_APPLICATIONS name, and the [ReactiveAppInstanceScanner#determineInstanceFromTargets](https://github.com/promregator/promregator/compare/master...tzolov:gh-35-dynamic-target-list?expand=1#diff-eac88078a44f3aaf65fdd4858abe390bR108) to call the [expandTargets](https://github.com/promregator/promregator/compare/master...tzolov:gh-35-dynamic-target-list?expand=1#diff-eac88078a44f3aaf65fdd4858abe390bR158) to add additional targets when the name is SPACE_ALL_APPLICATIONS

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
+++ b/src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
@@ -32,6 +32,8 @@ import io.prometheus.client.hotspot.DefaultExports;
 public class PromregatorApplication {
 	
 	private static final Logger log = Logger.getLogger(PromregatorApplication.class);
+
+	public static final String SPACE_ALL_APPLICATIONS = "SPACE_ALL_APPLICATIONS";
 	
 	public static void main(String[] args) {
 		SpringApplication.run(PromregatorApplication.class, args);
--- a/src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
+++ b/src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
@@ -16,6 +16,7 @@ import org.cloudfoundry.client.v2.shareddomains.SharedDomainEntity;
 import org.cloudfoundry.client.v2.spaces.SpaceResource;
 import org.cloudfoundry.client.v3.processes.ListProcessesResponse;
 import org.cloudfoundry.client.v3.processes.ProcessResource;
+import org.cloudfoundry.promregator.PromregatorApplication;
 import org.cloudfoundry.promregator.cfaccessor.CFAccessor;
 import org.cloudfoundry.promregator.config.Target;
 import org.cloudfoundry.promregator.internalmetrics.InternalMetrics;
@@ -30,13 +31,13 @@ import reactor.core.publisher.Mono;
 @Component
 public class ReactiveAppInstanceScanner implements AppInstanceScanner {
 	private static final Logger log = Logger.getLogger(ReactiveAppInstanceScanner.class);
-	
+
 	private PassiveExpiringMap<String, Mono<String>> orgMap;
 	private PassiveExpiringMap<String, Mono<String>> spaceMap;
 	private PassiveExpiringMap<String, Mono<String>> applicationMap;
 	private PassiveExpiringMap<String, Mono<String>> hostnameMap;
 	private PassiveExpiringMap<String, Mono<String>> domainMap;
-	
+
 	@Value("${cf.cache.timeout.application:300}")
 	private int timeoutCacheApplicationLevel;
 
@@ -48,20 +49,20 @@ public class ReactiveAppInstanceScanner implements AppInstanceScanner {
 
 	@Autowired
 	private CFAccessor cfAccessor;
-	
+
 	@Autowired
 	private InternalMetrics internalMetrics;
 
 	private static class InternalInstance implements Cloneable {
-		
+
 		public Target target;
-		
+
 		public Mono<String> orgId;
 		public Mono<String> spaceId;
 		public Mono<String> applicationId;
-		
+
 		public String instanceId;
-		
+
 		public String accessUrl;
 
 		@Override
@@ -71,19 +72,19 @@ public class ReactiveAppInstanceScanner implements AppInstanceScanner {
 			other.orgId = this.orgId;
 			other.spaceId = this.spaceId;
 			other.applicationId = this.applicationId;
-			
+
 			other.instanceId = this.instanceId;
-			
+
 			other.accessUrl = accessUrl;
 			return other;
 		}
-		
+
 		public Instance toInstance() {
 			return new Instance(this.target, this.instanceId, this.accessUrl);
 		}
-		
+
 	}
-	
+
 	@PostConstruct
 	public void setupMaps() {
 		this.orgMap = new PassiveExpiringMap<>(this.timeoutCacheOrgLevel, TimeUnit.SECONDS);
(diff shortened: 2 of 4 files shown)


src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
```diff
@@ -32,6 +32,8 @@
 public class PromregatorApplication {
 	
 	private static final Logger log = Logger.getLogger(PromregatorApplication.class);
+
+	public static final String SPACE_ALL_APPLICATIONS = "SPACE_ALL_APPLICATIONS";
```
eaglerainbow: see also my comment at https://github.com/promregator/promregator/issues/35#issuecomment-384415101 regarding the "magic identifier" topic.

eaglerainbow: As this constant is related only to `Target#applicationName`, I think it would be better to have that constant at `Target`.
This would also prevent that we have to `import org.cloudfoundry.promregator.PromregatorApplication` in `ReactiveCFAccessorImpl` and further classes too, which also violates the hierarchical class structure.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -95,75 +96,97 @@ public void setupMaps() {
...
 
+	private Flux<Target> expandTargets(Flux<Target> inputTargets) {
+
+		return inputTargets.flatMap(target -> {
+			if (!target.getApplicationName().equalsIgnoreCase(PromregatorApplication.SPACE_ALL_APPLICATIONS)) {
+				return Mono.just(target);
+			}
+			Mono<String> orgId = getOrgId(target.getOrgName());
+			Mono<String> spaceId = getSpaceId(orgId, target.getSpaceName());
+			Flux<ApplicationResource> apps = allApplicationByOrgAndSpace(orgId, spaceId);
```
eaglerainbow: I am not 100% sure, but if I get you right, then this is more or less a "pre-selection" using an own CF roundtrip. This leads to yet another roundtrip, which is used for resolving the wildcard identifier.

I am little skeptical here, as performance on the CF metadata API is quite crucial from what I saw from the performance tests conducted on promregator. I had tried to avoid as many roundtrips as I could - and (if I get it right) we are adding one. 
Perhaps remember: Scraping intervals are around 10-15s; with a full set of CF API requests, I measured that they may take up to 2s already (with only having two apps in the selection) for discovery only. And the discovery is on the critical path.

I fear that we won't get around to extend https://github.com/promregator/promregator/blob/ea9fc730ce19f8895f10d2213a441b3d47c2a029/src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java to handle this.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -306,12 +329,33 @@ public void stop() {
...
+					}
+
+					if (resources.isEmpty()) {
+						log.warn(String.format("Received empty result on requesting application %s",
+								PromregatorApplication.SPACE_ALL_APPLICATIONS));
+						return Flux.empty();
+					}
+
+					return Flux.fromIterable(resources);
+				}).filter(r -> r.getEntity().getState().equalsIgnoreCase("STARTED"));
```
eaglerainbow: That's an important aspect that we still need to cover in our discussion at #35 :  Apps in which state do we consider that we should discover for scraping? You have filtered on `STARTED` here. What about `STOPPED`? Are there further "bad states" (`CRASHED`?)?

BTW: Aren't there any constants interfaces for these literals?

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -95,75 +96,97 @@ public void setupMaps() {
...
+
+
 	public List<Instance> determineInstancesFromTargets(List<Target> targets) {
 		LinkedList<Instance> result = new LinkedList<>();
-		
-		Flux<Target> initialFlux = Flux.fromIterable(targets);
-		
-		Flux<InternalInstance> initialInstancesOnlyApplication = initialFlux.map( target -> {
+
+		Flux<Target> initialFlux = this.expandTargets(Flux.fromIterable(targets));
```
eaglerainbow: I am unsure if this is a great idea or not (perhaps you may comment on this from your point of view): Expanding the targets will be done on every call of `determineInstancesFromTargets`, which means on every request to scrape. As I assume that a pre-selection request will be fired (see also other code review comment), this would mean that this wildcard resolution bypasses the cache. 
Whilst this is a good idea from a consistency's point of view, it will be negative from the performance's / TCO's perspective. 

Was this decision pro consistency intentional?

eaglerainbow:
Oh, one more thing: If we make it live, we also have to add documentation to it (thus, I'll flag this PR also as documentation relevant)

src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
tzolov: +1

tzolov: Sure, using `*` character would help for a smother transition form filters based on static keys to a reg-ex like filters.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
tzolov: This is a relevant point.  It seems we would need a dedicated (target discovery) thread that might use a ReactiveCFAccessorImpl method to refresh the targets with refresh rate hopefully much slower than the metrics update rate  In addition the promregator can expose a endpoint to receive target-update signals (e.g. not receiving the new targets but just a signal that it should recompute its target wild cards)

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
tzolov: This relevant indeed. For my prototype i've pre-filtered only the STARTED apps but i  don't fill like this is correct generic behavior. We can either the state an yet another configurable filter or register all targets in spite of their sates but pass the state as dimensional tag along the measurements.  Not entirely sure how would this work with Prometheus though,.

eaglerainbow:
@tzolov I think mentally I have all the necessary pieces together to solve this puzzle here. 
Are you working actively on a modified version of this PR? If no, I would give it a try composing another proposal. My question is solely for the purpose of avoiding unnecessary double work.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
tzolov: I believe this overlaps with the other comments above, e.g. finding a efficient and less intrusive mechanisms to invalidate a and refresh the metadata?

tzolov:
@eaglerainbow feel free to go ahead and solve the puzzle! Let me know if i can help with something

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
eaglerainbow: I think we have to iterate on that a little. `STARTED` is correct for sure, but it most likely won't be the only one. Yet, `CRASHED` or `STOPPED` will not make sense to let through, as then also the scraping endpoint won't be available anyway. Even if it was, most likely nothing meaningful could come back. So, I would assume that Prometheus would want to see those instances with an
``` 
up{...} 0
```
anyway.
Making it configurable (in the special event that really some folks have deviating opinion) can also be done compatibly afterwards.

eaglerainbow:
@tzolov 

> feel free to go ahead and solve the puzzle! Let me know if i can help with something.

Thanks! Will try to do so. 
It would be great if you could perform a code-review once I have pushed my solution to and give comments!
(Don't expect that this will happen within less than 24h)

eaglerainbow:
@tzolov I am not fully satisfied with it yet (i.e. it's still work-in-progress), but if you want you may already have a first glimpse at https://github.com/promregator/promregator/tree/issue-35-applicationwildcard

PR #47, opened 2018-04-29
Discovery of all Applications within given Org/Space

eaglerainbow:
Issue #35's main request is to be able to detect applications within a CF space to prevent that you have to explicitly specify them individually. PR #46 provided a first implementation approach of doing this. The discussion of the PR has shown that it would have several limitations. Yet, the same discussion also showed ideas to prevent them. 

This PR tries to pick up these ideas and bring them into code form. The main principles are:
* Make the `applicationName` property in the target configuration optional
* If the `applicationName` is not specified, retrieve the entire list of applications within the specified space
* Each application in state `STARTED` is considered a valid target for scraping.
* The result of retrieving the entire list is cached with the same timeout as the application cache.

Furthermore, the following enhancements are provided:
* And HTTP REST endpoint which allows to invalidate the application cache (and other caches) immediately
* This HTTP REST endpoint can be auth-protected
* When reading the entire list of applications (as depicted above), the result is used for pre-filling the application cache, thus preventing another (new) roundtrip to be sent to the CF Cloud Controller.

## Credits
At least half of the credits for this PR goes to @tzolov, who did

1. a great PoC with #46 ,
2. contributed the main ideas on 
   * limiting the solution to only retrieving all applications in a space,
   * having a HTTP REST endpoint for cache invalidation, and
   * filtering on the state of CF apps during app discovery.

## Still to do after merging
- [ ] Add statement to release notes
- [ ] Close #46 as this PR supersedes it.


tzolov:
Hi @eaglerainbow, 
Sorry i couldn't do it earlier! Will try review and test the PR later today or earlier tomorrow.

eaglerainbow:
No worries! It's not the case that this is a total blocker for the world and all Promregators would be broken with a security flaw or so...

eaglerainbow:
@tzolov heads up for you: I just played around a little with a refactoring of the `ReactiveAppInstanceScanner` at branch [refactoringAppInstanceScanner](https://github.com/promregator/promregator/tree/refactoringAppInstanceScanner). This refactoring might be quite conflicting with this PR here -- and I am currently quite undecided, whether which one we should merge first. 

So, to safe-guard your personal time investment, perhaps for your review, primarily focus on the concepts and less on the coding...

eaglerainbow:
mmmh.. this enhancement here could have a bigger implication than we might think: With this coding enabled, it may happen that new applications are added and old one's are being dropped. Therefore, we must free up the resources for dropped applications accordingly, otherwise we will run into leak situations (especially I have the use cases depicted in #35 on my mind):

* The `PassiveExpiringMap`s at https://github.com/promregator/promregator/blob/eaf866993b8f4ae321789116aa1ed5c301e311e1/src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java are not a problem, as they are auto-cleaning themselves over the time.
* `InternalMetrics` at https://github.com/promregator/promregator/blob/eaf866993b8f4ae321789116aa1ed5c301e311e1/src/main/java/org/cloudfoundry/promregator/internalmetrics/InternalMetrics.java also are not of concern, as they don't record anything app-specific as of today (but in the future could so). Luckily, they are disabled by default: https://github.com/promregator/promregator/blob/eaf866993b8f4ae321789116aa1ed5c301e311e1/src/main/java/org/cloudfoundry/promregator/internalmetrics/InternalMetrics.java#L14
* On each scraping request a new instance of `requestRegistry` is created. https://github.com/promregator/promregator/blob/eaf866993b8f4ae321789116aa1ed5c301e311e1/src/main/java/org/cloudfoundry/promregator/endpoint/AbstractMetricsEndpoint.java#L84 Thus, that one only contains the metrics of those applications, which are currently subject to scraping ➡️ shouldn't be a problem either.
 
Seems like that we are lucky and by chance do not have this kind of issue.

eaglerainbow:
Note that there is no mechanism provided by the [`PassiveExpiringMap`](http://commons.apache.org/proper/commons-collections/javadocs/api-release/org/apache/commons/collections4/map/PassiveExpiringMap.html) as of today, which supports subscribing to expired values. 
Removal Listeners of https://github.com/google/guava/wiki/CachesExplained might come for rescue here.

eaglerainbow:
@tzolov given the problem with the possible memory leaks mentioned in my previous posts, it is now clear for me that the change in [refactoringAppInstanceScanner](https://github.com/promregator/promregator/tree/refactoringAppInstanceScanner) needs to go first. Otherwise, the refactoring of the cache handling can't happen properly (or, to be more precise, we would have to do that endeavor also twice, too). 
Thus, I will merge [refactoringAppInstanceScanner](https://github.com/promregator/promregator/tree/refactoringAppInstanceScanner) now, take the challenge on the caching problem afterwards. Once the latter is finished, I'll come back to this PR again...

For you this means: you may still look at it, but be prepared that the coding will still change a second time significantly. So, don't dig too deep into it, yet.

eaglerainbow:
@tzolov It's good that you did not add comments to this PR, yet, as I had to re-implement it almost in its entirety with PR #53.  Therefore, this PR is now in a stale state.

PR #53 isn't fully completed, yet (see also notes there), but already in a very mature state. Therefore, if you want to already have a look at it, it's a good time to do so.
Thanks!

PR #48, opened 2018-05-02
Filter Targets to be Scraped based on Application Names using a Regular Expression

eaglerainbow:
Following up PR #47, this PR enriches the configuration options to filter application names by a regular expression. The regular expression can be specified in attribute `applicationRegex` of a target. 

## Hint and Warning
Please only start reviewing this PR **after** #47 has been merged (as this PR was built on top of #47 and thus github currently shows the delta between `master` and this PR and not just between #47 and this PR). 

## Still to do after merging
- [ ] Add statement to release notes

PR #53, opened 2018-05-04
Discovery of all Applications within a given Org/Space

eaglerainbow:
This is a reimplementation of PR #47 .

During the development of #47 it turned out that it is necessary to de-register metrics from the `CollectorRegistry` to prevent a resource leakage in case we want to allow auto-detection of (all) applications with in a given CF Org/Space (original request with #35).  To allow that, on the current `master` codeline many preparations have already taken place:

* 26c5f024e5b105b1b1312a52f593ca806627152b introduces the concept of `TargetResolver`s. 
* Up to 9076448557be67bfc5a48c410121c89fafaa6ef1 most of the caches, which had been in `ReactiveAppInstanceScanner` before, have been pushed down (and thus centralized) to the `ReactiveCFAccessorImpl`.

Especially the concept of `TargetResolver`s allows a complete different structure to implement the requirement of discovering applications within given (fixed) orgs/spaces.

The main principles for implementation are still unchanged, though:
* Make the `applicationName` property in the target configuration optional
* If the `applicationName` is not specified, retrieve the entire list of applications within the specified space
* Each application in state `STARTED` is considered a valid target for scraping.
* The result of retrieving the entire list is cached with the same timeout as the application cache.

On top, the following enhancements are provided:
* And HTTP REST endpoint which allows to invalidate the application cache (and other caches) immediately
* This HTTP REST endpoint can be auth-protected
* When reading the entire list of applications (as depicted above), the result is used for pre-filling the application cache, thus preventing another (new) roundtrip to be sent to the CF Cloud Controller.

## Credits
At least half of the credits for this PR goes to @tzolov, who did

1. a great PoC with #46 ,
2. contributed the main ideas on 
   * limiting the solution to only retrieving all applications in a space,
   * having a HTTP REST endpoint for cache invalidation, and
   * filtering on the state of CF apps during app discovery.

## Still to do after merging
- [x] Add statement to release notes
- [ ] Close #46 as this PR supersedes it. 
- [ ] Close #47 as this PR supersedes it. 
- [ ] Reimplement #48 as this PR is stale as soon as this PR is merged.


eaglerainbow:
After thinking about the current approach in this PR, I slowly get some doubts, if that is the right approach:
* The lifecycle between the `CachingTargetResolver` and the `MetricsFetcherMetrics` is totally different: While the `CachingTargetResolver` is an ever-lasting singleton spring bean, the  `MetricsFetcherMetrics` has only a lifetime of the a single request (i.e. as long as a `AbstractMetricsEndpoint` is living). Multiple requests will also create multiple instances of `MetricsFetcherMetrics`. The metrics a `MetricsFetcherMetrics` generates are registered at the default registry of the Prometheus Client, though. Thus, it is dangerous to register each instance of `MetricsFetcherMetrics` to the `CachingTargetResolver` as `RemovalListener`. Over time, we could get hundreds of them - for requests, which already have died out anyway.
* Registering them via a weak reference also does not help: If nobody would be scraping, then we would no such instance of a `MetricsFetcherMetrics` at hand. Thus, we also would have no listener, which could deregister the samples (which might be necesary).

This all would suggest that we should have a 1:1 relationship between a `TargetRemover` and the `ResolvedTarget`. However, 

* Removal from the cache at the `CachingTargetResolver` does not necessarily mean that the application is invalid on the platform! E.g. the cache may displace the target as it has expired, but on next occasion of scanning, the target would be re-added to the cache, as it is still valid. If in this case the removal would also deregister the sample/metric, then the samples' data would be lost and the counters would be reset to 0. This may happen quite often (e.g. every 5 minutes), which would make the exposed metrics almost useless.

In short, this is a bigger lifecycle problem here to ensure that our caches are efficient, but neither stale. Moreover, the metrics of `MetricsFetcherMetrics` have a relationship of a composition to the application on the platform; but this is not reflected in our design, yet...

eaglerainbow:
> After thinking about the current approach in this PR, I slowly get some doubts

With 41a92ff0b6ea846825013da17b58ea190d8e559f I now did a bigger round on this entire problem and merged the `InstanceLifecycleHandler`. In a nutshell, the `InstanceLifecycleHandler` in combination with the `CFDiscoverer` now takes care about an own timeout (typically: 10 minutes), after which an instance is declared invalid. However, that invalidation may be prolonged as many times as possible. Only after the timeout, via a JMX/AMQ (which allows a proper decoupling of the instances), with 09ec7d87db02f8d0c2281364f2c2ca51ff06f945 the `MetricsFetcherMetrics` is called to `deregisterSamplesFromRegistry`. That solves our lifecycle problem stated above.

eaglerainbow:
@tzolov: After a longer battle with this PR (and many further preparations made on `master`), I think I finally got this one right now. At least no question/issue is open to my mind anymore. 

If you are still interested in reviewing this PR, please feel free to do so and briefly leave a comment here. If not (which I could also totally understand!), I would merge this PR - let's say - on May, 23rd (if you leave a comment to review, I wouldn't be doing that to prevent putting you under time pressure).

eaglerainbow:
Ok, seems like @tzolov didn't find time  to look at this anymore (that's ok!).
I'll go and merge this beast here now...

PR #55, opened 2018-05-11
Adapt label block regex to parse values with curly braces

vervas:


eaglerainbow:
I like the idea!

eaglerainbow:
@vervas Do you have a problem with this case right now on 0.3.2?

vervas:
Yes, because many of our metrics contain curly braces thus we lose certain series because of this. I've built a new image from my machine and used that one, however whenever I deploy something I've built on my local machine, it goes crazy on CPU usage and eventually dies (this happens either with this change or without). Am I missing something when building with maven?

eaglerainbow:
> Yes, ...

I have released [0.3.3](https://github.com/promregator/promregator/releases/tag/v0.3.3) for you...

> deploy something I've built on my local machine, it goes crazy on CPU usage and eventually dies [...] Am I missing something when building with maven?

Not that I would be aware of something special. Everything which I have in the build process of the Docker container is written to the `Dockerfile`. 
The only thing that is "special" is the Docker container in which I build the docker image - but this is mainly due to the fact that my CI environment is a little special....

Which JDK version do you use? 
On the build machine, there is currently running OpenJDK 1.8.0_151 amd64 and Apache Maven 3.3.9

PR #88, opened 2019-02-08
Enable encrypted properties in config files.

nicolaimainiero:
To enable encrypted properties as decribed here: https://www.novatec-gmbh.de/en/blog/encrypted-properties-spring/ I made the following changes.

- add spring-cloud-starter dependency
- update spring-cloud.version to Finchley.SR2

### DIFF
--- a/docs/passwords-in-config.md
+++ b/docs/passwords-in-config.md
@@ -76,3 +76,10 @@ promregator:
         password: '{cipher}d93dd725143e187804e4105a681706c1185c58f92086bf55aa830e61c0527060'
 ```
 
+If you are running promregator as docker container you can provide the environment variable in the following way
+```
+docker run -d \
+ --env ENCRYPT_KEY=mySecretKey \
+ -v /path/to/your/config/with/encrypted/attributes/myconfig.yaml:/etc/promregator/promregator.yml \
+ promregator/promregator:<version>
+```
--- a/docs/passwords-in-config.md
+++ b/docs/passwords-in-config.md
@@ -61,7 +61,7 @@ d93dd725143e187804e4105a681706c1185c58f92086bf55aa830e61c0527060
 Then you need to set the encryption key as environment variable so that promregator can process it:
 ```bash
 export ENCRYPT_KEY=mySecretKey
-java -Dspring.config.location=file:/path/to/your/config/with/encrypted/attributes/myconfig.yaml -jar promregator-0.0.1-SNAPSHOT.jar
+java -Dspring.config.location=file:/path/to/your/config/with/encrypted/attributes/myconfig.yaml -jar promregator-x.y.z-SNAPSHOT.jar
 ```
 
 The config file would the look something like this. Note the value of the password attribute.
--- a/docs/passwords-in-config.md
+++ b/docs/passwords-in-config.md
@@ -30,7 +30,7 @@ In the example above you see that the password was not specified directly, but a
 
 ```bash
 export MY_PASSWORD_OS_VARIABLE=mysecretPassword
-java -Dspring.config.location=file:/path/to/your/myconfig.yaml -jar promregator-0.0.1-SNAPSHOT.jar
+java -Dspring.config.location=file:/path/to/your/myconfig.yaml -jar promregator-x.y.z.jar
 ```
 
 Observe here that the placeholder `${my.password.os.variable}` gets capitalized and all dots in the placeholder gets replaced by the underscore symbol to form the OS environment variable you have to set.
@@ -41,7 +41,7 @@ Note that you could also set the Java system properties with the same naming con
 ## Note on Setting OS Environment Variables using Docker
 Keep in mind that you able to set OS environment variables for the Docker container by using 
 ```bash
-docker run --env MY_PASSWORD_OS_VARIABLE=mysecretPassword promregator/promregator:0.0.1
+docker run --env MY_PASSWORD_OS_VARIABLE=mysecretPassword promregator/promregator:x.y.z
 ```
 
 Further information on this can also be found in the official [Docker reference documentation](https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file).
--- a/docs/passwords-in-config.md
+++ b/docs/passwords-in-config.md
@@ -48,10 +48,23 @@ Further information on this can also be found in the official [Docker reference 
 
 ## Using Encrypted Password in Configuration
 
-As an alternative for using environment variables for the password you can use encrypted password (values starting with `{cipher}`) in the configuration.
-See also the documentation about [Encryption & Decryprtion](http://cloud.spring.io/spring-cloud-config/spring-cloud-config.html#_encryption_and_decryption) of the Spring Cloud Config Server.
+As an alternative for using environment variables for the password you can use encrypted password (values starting with `{cipher}`) in the configuration. 
+This allows you to check the configuration into a version control system. 
+See also the documentation about [Encryption & Decryption](http://cloud.spring.io/spring-cloud-config/spring-cloud-config.html#_encryption_and_decryption) of the Spring Cloud Config Server.
 
-Following example illustrates the usage:
+To encrypt a value you can use the [Spring Cloud CLI](https://cloud.spring.io/spring-cloud-cli/) extension and encrypt the password on the command line.
+```bash
+spring encrypt mysecret --key mySecretKey
+d93dd725143e187804e4105a681706c1185c58f92086bf55aa830e61c0527060
+```
+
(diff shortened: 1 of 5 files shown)


eaglerainbow:
Thanks for sharing this. I will have a look at it...

eaglerainbow:
Please ignore the issue with the failing voter; I will take care of this (you don't stand a chance, as the server is within my intranet).

nicolaimainiero:
Hi @eaglerainbow, I finally managed to update the docs on encrypted passwords. Sorry it took so long.
Regarding the integration test, do you have already an idea how to integrate them?

docs/passwords-in-config.md
```diff
@@ -46,3 +46,37 @@ docker run --env MY_PASSWORD_OS_VARIABLE=mysecretPassword promregator/promregato
...
+``` 
+* Use the Spring Cloud CLI extension and encrypt the password on the command line.
+```bash
+spring encrypt mysecret --key foo
+682bc583f4641835fa2db009355293665d2647dade3375c0ee201de2a49f7bda
+```
+
+You still need to provide the decryption key to your application for example as environment variable:
+```bash
+export ENCRYPT_KEY=mySecretKey
```
eaglerainbow: This comes a little late for me. That is to say that (as far as I do understand it), setting this environment variable is also necessary for the two cases above (using the `/encrypt` endpoint - then the server must be started with this environment variable - and7or the Spring Cloud CLI).
Thus, we should mention this a little earlier.

docs/passwords-in-config.md
```diff
@@ -46,3 +46,37 @@ docker run --env MY_PASSWORD_OS_VARIABLE=mysecretPassword promregator/promregato
...
+java -Dspring.config.location=file:/path/to/your/enrypted/myconfig.yaml -jar promregator-0.0.1-SNAPSHOT.jar
```
eaglerainbow:
```suggestion
java -Dspring.config.location=file:/path/to/your/encrypted/myconfig.yaml -jar promregator-0.0.1-SNAPSHOT.jar
```

eaglerainbow: perhaps mentioning 0.0.1 explicitly here isn't such a great idea. Either we should write `x.y.z` or we should take the most current version

docs/passwords-in-config.md
```diff
@@ -46,3 +46,37 @@ docker run --env MY_PASSWORD_OS_VARIABLE=mysecretPassword promregator/promregato
...
+      type: basic
+      basic: 
+        username: username
+        password: '{cipher}682bc583f4641835fa2db009355293665d2647dade3375c0ee201de2a49f7bda'
+```
+
+To create the encrypted password you have two options. 
+* Use a Spring Cloud Config Server and POST the password to the `/encrypt` endpoint
+```bash
+curl localhost:8888/encrypt -d mysecret
```
eaglerainbow: Did you check whether this worked also with Promregator already out of the box? I hope not, because then we might be subject to a chosen-ciphertext attack...

docs/passwords-in-config.md
```diff
@@ -46,3 +46,37 @@ docker run --env MY_PASSWORD_OS_VARIABLE=mysecretPassword promregator/promregato
...
+682bc583f4641835fa2db009355293665d2647dade3375c0ee201de2a49f7bda
```
eaglerainbow: I think that should still go with the previous line (and should not be written on a line of its own)...

docs/passwords-in-config.md
```diff
@@ -46,3 +46,37 @@ docker run --env MY_PASSWORD_OS_VARIABLE=mysecretPassword promregator/promregato
...
+``` 
+* Use the Spring Cloud CLI extension and encrypt the password on the command line.
+```bash
+spring encrypt mysecret --key foo
+682bc583f4641835fa2db009355293665d2647dade3375c0ee201de2a49f7bda
```
eaglerainbow: Same as above: should still go to the old line, right?

docs/passwords-in-config.md
nicolaimainiero: Changed it to x.y.z, otherwise this needs to be updated after every release.

nicolaimainiero: Changed it according to your suggestion.

nicolaimainiero:
So I tried to reword and make it more clear. Removed the instructions for Cloud Config Server, since this is not the focus of this PR.
Extended an unit test to validate if the encryption is working.

src/test/java/org/cloudfoundry/promregator/SpringBootLoadPropertiesForTestingSpringApplication.java
```diff
@@ -21,4 +24,8 @@ public Object anything() {
 	public void check() {
 		Assert.assertTrue(dummyValue);
 	}
+
+	public String getEncryptedValue() {
```
eaglerainbow: Sorry for sounding nit-picky: `getEncryptedValue` sounds to me as if the value I get back is still encrypted 😀 (well, my brain might be biased having studied cryptography some years back...). Naming it `getPlaintextValue()` might be more correct (it's what cryptography would say), but could also be confusing - especially, if we keep the attributes name as `encryptedValue`, which would awkwardly look strange:

```java
public String getPlaintextValue() {
    return encryptedValue;
}
```
😨 

In fact the whole story about this `'{cipher}...'` thing is that it should be transparent to the `SpringConfiguration`.

What do you think about changing the spring property's name from `encrypted.value` to `secret.value` (not describing how it was done, but for what it - typically - is used) and then calling this method `getSecretValue()` here? IMO this would avoid this weird discussion altogether.

docs/passwords-in-config.md
```diff
@@ -46,3 +46,33 @@ docker run --env MY_PASSWORD_OS_VARIABLE=mysecretPassword promregator/promregato
...
+
+To encrypt a value you can use the [Spring Cloud CLI](https://cloud.spring.io/spring-cloud-cli/) extension and encrypt the password on the command line.
+```bash
+spring encrypt mysecret --key mySecretKey
+d93dd725143e187804e4105a681706c1185c58f92086bf55aa830e61c0527060
+```
+
+Then you need to set the encryption key as environment variable so that promregator can process it:
+```bash
+export ENCRYPT_KEY=mySecretKey
```
eaglerainbow: Do you know how to work with docker?
If yes, I think we should mention at least in a one-liner that users making use of our docker image may set this environment variable by using the `--env` parameter (see also one section above). Moreover, we also should consider to enable making use of [Docker Secrets](https://docs.docker.com/engine/swarm/secrets/) here as well (we can do so here, as the name of the environment variable is static and known at build time of the image).

If no, please just ignore my comment (and I'll create a separate issue for that).

eaglerainbow:
Just for the record: I tested with 9ee5ccd2d2a9f8095070b3ac11deaa769d6651b1, having set an `ENCRYPT_KEY` environment variable sending
```bash
$ curl http://localhost:8080/encrypt -d dummy
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   122    0   117  100     5    117      5  0:00:01 --:--:--  0:00:01  114k{"timestamp":"2019-05-04T13:04:04.506+0000","status":403,"error":"Forbidden","message":"Forbidden","path":"/encrypt"}
```
Access to the `/encrypt` endpoint is forbidden by default 😓 .

src/test/java/org/cloudfoundry/promregator/SpringBootLoadPropertiesForTestingSpringApplication.java
nicolaimainiero: It's not nitpicking. I appreciate the clear idea of how the code should be written. I changed it to `getSecretValue`.

docs/passwords-in-config.md
nicolaimainiero: I added the docker insturctions. Unfortunately, I have no experience with Docker Secrets, so I can't help you with this part.

eaglerainbow:
I just tested this feature manually myself. It looks good.
However, you should know that there is an incompatibility on the spring-boot-cli at 2.0.9+ and above against the spring-boot-cloud-cli 2.0.0. You *must* use spring-boot-cli:2.0.8 right now. 

@nicolaimainiero I am a little unsure whether we should mention this issue in our documentation or not. Are you aware of any official documentation on this where we could link to? I had to google for it and found some corresponding comments at SO, which then helped me out...

eaglerainbow:
Besides this little question, I would be willing to merge this PR.

eaglerainbow:
I added a comment on the version clash with 815df19ac66761a2967a6d2065af58268c0859e9.

PR #95, opened 2019-04-16
Update singleTargetScraping.md

huningd:
I think you mean advantage instead of disadvantages


eaglerainbow:
You have a point in there 😁 
Thanks for contributing!

PR #125, opened 2019-11-23
CVE-2019-12402 & JSON Injection: Feature/security fixes

warlockza:
Fixes to resolve vulnerability in commons-compress
Fixes for potential Json injection identified by Fortify

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,7 +38,7 @@
 	<parent>
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
-		<version>2.1.9.RELEASE</version> <!-- see also https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-parent -->
+		<version>2.1.10.RELEASE</version> <!-- see also https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-parent -->
 		<relativePath /> <!-- lookup parent from repository -->
 	</parent>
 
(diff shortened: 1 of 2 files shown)


eaglerainbow:
First of all, thanks for contributing and raising this issue here!
Please (also) see my comments on your changed lines.

pom.xml
```diff
@@ -128,13 +128,25 @@
 			<artifactId>cloudfoundry-client-reactor</artifactId>
 			<version>${cloudfoundry.client.version}</version>
 		</dependency>
+		<!--Need to override version of compress due to CVE-2019-12402 -->
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-compress</artifactId>
+            <version>1.19</version>
+		</dependency>
```
eaglerainbow: I am in general not in favor of changing versions of transitive dependency - often, it's asking for trouble. The only exemption to the rule is security (which you have pulled here 😁 )

The dependency is coming from cloudfoundry/cf-java-client (here: cloudfoundry-client-reactor:3.15.0.RELEASE`). The version we are currently using has this dependency to `commons-compress:1.18`.
cloudfoundry/cf-java-client has released [3.17.1.RELEASE in the meantime](https://github.com/cloudfoundry/cf-java-client/releases/tag/v3.17.1.RELEASE). That uses `commons-compress:1.19` out of the box. So, we wouldn't have to provide a manual overwrite of the version. Hence, I am currently more in favor of bumping the version of the entire artifact. What do you think?

warlockza: I would bump the entire artifact , that is cleanest. My only concern was any subtle behaviour changes between cf client versions, which is why I chose to override as the safest solution.

eaglerainbow: > My only concern was any subtle behaviour changes between cf client versions

I do understand your concern, which I share with you in general. However, in this special case, there are two reasons for me why I think it's not a problem:
1. Compatibility of cloudfoundry/cf-java-client (except for bugs 😀) is really good so far.
2. We are talking here about `master`. We cannot dig ourselves in to a single version: Not bumping now would mean that we have to fear bumping also in the feature. That's not acceptable.

I have added efe9e52caa9516f942b237480684cdf43f6ca48a which bumps the surrounding library version.

pom.xml
```diff
@@ -128,13 +128,25 @@
 			<artifactId>cloudfoundry-client-reactor</artifactId>
 			<version>${cloudfoundry.client.version}</version>
 		</dependency>
+		<!--Need to override version of compress due to CVE-2019-12402 -->
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-compress</artifactId>
+            <version>1.19</version>
```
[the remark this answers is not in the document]
eaglerainbow: became obsolete.

eaglerainbow:
Due to [this statement](https://github.com/promregator/promregator/issues/126#issuecomment-557845930) equivalent functionality merged with efe9e52caa9516f942b237480684cdf43f6ca48a and  f0b592ccf185de23e6b7380e7628b36424f322ca also crediting the original author. 

Thanks for reporting this!

Closing this PR, as no aspects is missing anymore in the mainline.

PR #129, opened 2019-11-30
Fix to AutoRefreshingCacheMapTest.java to be more Tolerant of slowness

warlockza:


src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
```diff
@@ -86,8 +89,7 @@ public void testAutoRefresh() throws InterruptedException {
 		Assert.assertEquals("initial", value);
 		
 		// we have to wait until the thread had a chance to refresh
-		Thread.sleep(1000);
-		
+		await().atMost(5, SECONDS).until(() -> !subject.get("key").equals("initial"));
```
eaglerainbow: I fear this changes the semantics of the test: `subject.get()` would load the entry, if it was evicted. See also https://github.com/promregator/promregator/blob/744b4044fc424c3ba729e7c354a66f2c7c328589/src/main/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMap.java#L147 

Note that `subject.containsKey()` does not this.
However, what we want to test here

eaglerainbow:
Could

```java  
subject.setRefresherThreadWithIncreasedPriority(true);
```

help here?

src/main/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMap.java
```diff
@@ -66,7 +66,7 @@ public Object getLockObject() {
 		
 	}
 	
-	private Map<K, EntryProperties> entryPropertiesMap = Collections.synchronizedMap(new HashMap<>());
+	private final Map<K, EntryProperties> entryPropertiesMap = Collections.synchronizedMap(new HashMap<>());
```
warlockza: This was also a Sonar Recommendation as we are synchronizing on it

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
```diff
@@ -62,33 +65,36 @@ public void testExpiry() throws InterruptedException {
...
+
+		subject.put(testKey, "initial");
+		String value = subject.get(testKey);
 		Assert.assertEquals("initial", value);
 		
 		// we have to wait until the thread had a chance to refresh
-		Thread.sleep(1000);
-		
-		boolean exists = subject.containsKey("key");
+		await().atMost(5, SECONDS).until(() -> counter.getCounter()>0);
```
warlockza: I had a modified version of this test that did 500 iterations and timed for outliers. Only with these changes did we actually complete 500 iterations without failing the assert (ie not refreshing). Most outliers were 1001-1060ms on my desktop, but i have seen outliers of  +/- 1500ms.

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
warlockza: Hopefully my updated changes do not change test semantics

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
```diff
@@ -265,13 +269,15 @@ public boolean equals(Object obj) {
 			if (a == null) {
 				if (other.a != null)
 					return false;
-			} else if (!a.equals(other.a))
+			} else if (!a.equals(other.a)) {
```
eaglerainbow: Just for the record: I understand what's up here, but we will not have fun with it: These kind of methods are auto-generated by Eclipse.... So, as soon as we regenerate them, we have to fix the SonarQube Recommendation again.

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
```diff
@@ -62,37 +65,38 @@ public void testExpiry() throws InterruptedException {
...
 		
 		// we have to wait until the thread had a chance to refresh
-		Thread.sleep(1000);
-		
-		boolean exists = subject.containsKey("key");
-		Assert.assertTrue(exists);
-		
-		Assert.assertEquals("refreshed", subject.get("key"));
-		Assert.assertEquals(1, counter.getCounter());
+		await().atMost(3, SECONDS).untilAsserted(() -> Assert.assertEquals(1, counter.getCounter()));
```
eaglerainbow: Good idea.
But why so complicated? Wouldn't 

```java
await().atMost(3, SECONDS).until(() -> counter.getCounter() >= 1));
``` 
also work?
At least this would be much easier to understand for me.

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
```diff
@@ -62,37 +65,38 @@ public void testExpiry() throws InterruptedException {
...
 		}
 	}
-	
+
 	@Test
-	public void testAutoRefresh() throws InterruptedException {
+	public void testAutoRefresh() {
 		final Counter counter = new Counter();
-		
+		String testKey="key";
```
eaglerainbow:
```suggestion
		final String testKey = "key";
```

src/main/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMap.java
```diff
@@ -255,7 +255,7 @@ public void run() {
 			MDC.remove("AutoRefreshingCacheMap");
 		}
 		
-		private void refreshMap() {
+		private synchronized void refreshMap() {
```
eaglerainbow: Why do you want to make it `synchronized` here?
The only one calling it is line 240, and per CacheMap there can only be exactly one thread running.

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
warlockza: We were doing an assert on counter being one anyway. So by doing the await for the counter to be 1 , and then the later assert on it.. its kind of wasteful. This gets the assert and the 'condtion' done in one line

src/main/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMap.java
warlockza: Happy to actually revert this change now ...

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
warlockza: Does Eclipse have updated templates that dont upset Sonar?

src/main/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMap.java
warlockza: Reverted

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
eaglerainbow: Not updated, but you may configure it: When generating set the flag "use blocks in 'if' statements".

Sounds like something that we should write down in some sort of project style guidelines...(I will take care)

warlockza: How do you wish to proceed?

eaglerainbow: done with cd744f0503c3660621b968bc0f71f3edba6cda95

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
eaglerainbow: okay, you convinced me.

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
eaglerainbow: > How do you wish to proceed?

Let's merge your stuff. The [new style guide](https://github.com/promregator/promregator/blob/81078620d64fd33bdad1255acfa08f4b2e528403/CONTRIBUTING.md) should make sure that future usage of the Eclipse generator will not break it anymore.

PR #130, opened 2019-11-30
Sonar Fixes for squid:S2293 and removed unused variables

warlockza:
It would seem that Master is failing with these 3 failures

Failures: 
  TextFormat004ParserTest.testIssue104:839 expected:<1> but was:<3>
  TextFormat004ParserTest.testVariant1:811 Values should be different. Actual: UNTYPED
  MergableMetricFamilySamplesTest.testIssue104:113 expected:<50> but was:<55>

I fail with the same 3 so from what I can tell I have not broken anything :)

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMap.java
+++ b/src/main/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMap.java
@@ -66,11 +66,11 @@ public class AutoRefreshingCacheMap<K, V> extends AbstractMapDecorator<K, V> {
 		
 	}
 	
-	private Map<K, EntryProperties> entryPropertiesMap = Collections.synchronizedMap(new HashMap<K, EntryProperties>());
+	private Map<K, EntryProperties> entryPropertiesMap = Collections.synchronizedMap(new HashMap<>());
 	private String name;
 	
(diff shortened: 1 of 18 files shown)


eaglerainbow:
> It would seem that Master is failing with these 3 failures

I am confused: I checked out both 6d9239361d76688567a0f927210afe9e9711f0a2 (a.k.a. current `master`) and your 3c9be240974c3023cf33c7123402e6f7ab29aaea. For both of them the unit tests run through just like charm (even the "beloved" `testMassRun`). 

Do you have anything special with your hardware setup?

warlockza:
Not particularly. I run it with Intellij. It does run the tests in parallel rather than serially so maybe some kind of concurrency issue?

eaglerainbow:
That could be the case: I have many unit tests which run in parallel (which also causes some SonarQube-detecting issues I just learned). It may be that running them in parallel on top may overload your machine. 
I don't know IntelliJ enough: Is it possible to test-run them serially?

warlockza:
I found out why they were failing. The Line Seperators for some reason are/were crlf not cr. so the parser was failing. What is the project default line separator, because it does appear to be inconsistent across some files in the project? I'm thinking it should be 'lf' ?

eaglerainbow:
> What is the project default line separator, because it does appear to be inconsistent across some files in the project? I'm thinking it should be 'lf' ?

We are unix-ish: So default is 'lf', if not a tool explicitly requires otherwise.
Where did you stumble over them? Perhaps it's even wrong in the master repo? If that is the case, we should fix it...

PR #131, opened 2019-11-30
Sonar Fixes for squid:S4488 and fixed some spelling

warlockza:


src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
```diff
@@ -218,7 +218,7 @@ public void forceGC() {
 	 * @return the unique identifier of the currently running instance
 	 */
 	@Bean
-	public UUID promregatorInstanceIdentifer() {
+	public UUID promregatorInstanceIdentifier() {
```
eaglerainbow: What a luck that `@Autowired` also works on type basis.
(nothing to be changed here!)

src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
```diff
@@ -167,7 +167,7 @@ public InternalMetrics internalMetrics() {
 	
 	@Value("${promregator.endpoint.threads:#{null}}")
 	@Deprecated
-	/**
+	/*
```
eaglerainbow: I am a little unsure whether removing the javadoc aspect here is the right thing. In Eclipse, it's no longer shown what shall be used instead...

src/main/java/org/cloudfoundry/promregator/endpoint/AbstractMetricsEndpoint.java
```diff
@@ -86,7 +86,7 @@
 
 	@Value("${promregator.endpoint.maxProcessingTime:#{null}}")
 	@Deprecated
-	/**
+	/*
```
eaglerainbow: I am a little unsure whether removing the javadoc aspect here is the right thing. In Eclipse, it's no longer shown what shall be used instead...

src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
warlockza: Its either a javadoc or a comment.. If its a javadoc, it should be fully javadocced .. if its just a comment, we treat it as a comment. My understanding of the issue above is that its currently half javadoc, half comment.

so either
/*  
xxxx
*/

or 

/**
Javadockystuff
**/

src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
```diff
@@ -167,7 +167,7 @@ public InternalMetrics internalMetrics() {
...
 	 * use threadPoolSize instead
```
eaglerainbow:
```suggestion
	 * The number of threads of the scraping thread pool. The value is coming from the deprecated configuration option <pre>promregator.endpoint.threads</pre>. Use threadPoolSize instead
```

src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
eaglerainbow:
```suggestion
	/**
```
Then let's make it to a real JavaDoc

src/main/java/org/cloudfoundry/promregator/endpoint/AbstractMetricsEndpoint.java
eaglerainbow:
```suggestion
	/**
```

src/main/java/org/cloudfoundry/promregator/endpoint/AbstractMetricsEndpoint.java
```diff
@@ -86,7 +86,7 @@
...
 	 * use maxProcessingTime instead
```
eaglerainbow:
```suggestion
	 * The maximal processing time permitted for Scraping (in milliseconds). The value is deprecated as it originates from the deprecated configuration option <pre>promregator.endpoint.maxProcessingTime</pre>. Use maxProcessingTime instead.
```

warlockza: So I worked out what a dangling Javadoc is now

Its where the javadoc is not completely above the property/method and its annotations.

Apparently, it needs to be before any annotations..... updated...

eaglerainbow:
BTW: You are aware of https://sonarcloud.io/dashboard?id=promregator ?
Just managed to configure this today...

warlockza:
Yes - I had a thorough peek at it . i saw you spent a good chunk of time cleaning stuff up.  I'm picking the low hanging non-invasive stuff until I am more familiar with the code base.

eaglerainbow:
Sounds good - but: I feel a little ashamed that you would have to do the nasty task and be the "vacuum cleaner" for my sloppiness 😁 
So, please feel encourage to also "really touch" the coding once you feel comfortable doing so!

PR #133, opened 2019-11-30
Fixed squid:S1602 and squid:S1612

warlockza:


warlockza:
Fixed Merge Conflicts with master

eaglerainbow:
On the changes on `ReactiveCFAccessorImpl.java`:
I saw the recommendations by SonarQube, too, but I was very unsure, whether we should adopt them or not. My main concern here is that it is intention to make the coding simpler and easier to read. While this definitively holds true for cases such as https://github.com/promregator/promregator/pull/133/files#diff-db7c28fb330982396a98a451a89314bbR186 and https://github.com/promregator/promregator/pull/133/files#diff-db7c28fb330982396a98a451a89314bbR271, I am not so convinced at cases such as https://github.com/promregator/promregator/pull/133/files#diff-cd96d2c1b0fda9ca65f7a8920bd54f90R275. Especially with the latter case the "important thing almost gets lost" - and that's `this.cloudFoundryClient.organizations().list(or)`. Also with https://github.com/promregator/promregator/pull/133/files#diff-db7c28fb330982396a98a451a89314bbR364 the side-effect procedural character is hidden behind a functional facade.

eaglerainbow:
What do you think? Make these changes really the code easier to read and understand?

warlockza:
Its a matter of personal preference. We can obviously put line breaks in to draw attention to things or to make things clearer. but on the whole it makes sense (at least  to me)

eaglerainbow:
> but on the whole it makes sense (at least to me)

Okay, then I am convinced.

PR #134, opened 2019-12-01
Fixing squid:S1488 and squid:S3973

warlockza:


src/main/java/org/cloudfoundry/promregator/cfaccessor/CacheKeyApplication.java
```diff
@@ -59,18 +59,21 @@ public boolean equals(Object obj) {
 		if (applicationName == null) {
 			if (other.applicationName != null)
 				return false;
-		} else if (!applicationName.equals(other.applicationName))
+		} else if (!applicationName.equals(other.applicationName)) {
```
eaglerainbow: see also https://github.com/promregator/promregator/pull/129/files#r352367615

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheClassic.java
```diff
@@ -1,396 +1,392 @@
-package org.cloudfoundry.promregator.cfaccessor;
```
eaglerainbow: You converted the entire file from newline format `lf` (which is the default for Promregator) to `cflf`. This isn't good 😬

warlockza: Yes, have been having some wierd issues. Should be fixed now

eaglerainbow:
BTW: CI job is just running (I have to still trigger them manually - many reasons...).
We should see an update in [SonarCloud](https://sonarcloud.io/dashboard?id=promregator) soon.

eaglerainbow:
Finished....
Fixing the old one's caused new Sonar violations to appear...

PR #136, opened 2019-12-01
Fixed squid:S2142

warlockza:


src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
```diff
@@ -133,6 +133,7 @@ public void run() {
 				Assert.assertEquals("labolg", value);
 			} catch (InterruptedException e) {
 				Assert.fail("Thread was interrupted");
+				Thread.currentThread().interrupt();
```
eaglerainbow: Well, won't be reached, as `Assert.fail` never returns. 
But Sonar doesn't know this...

Shall I put a "false positive" to SonarCloud on these cases?

src/test/java/org/cloudfoundry/promregator/mockServer/DefaultMetricsEndpointHttpHandler.java
```diff
@@ -29,6 +29,7 @@ public void handle(HttpExchange he) throws IOException {
 				Thread.sleep(this.delayInMillis);
 			} catch (InterruptedException e) {
 				Assert.fail("Unexpected interruption of mocking environment");
+				Thread.currentThread().interrupt();
```
eaglerainbow: see above

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
```diff
@@ -200,6 +201,7 @@ public void testLockObjectStringWorks() throws InterruptedException {
 				Thread.sleep(200);
 			} catch (InterruptedException e) {
 				Assert.fail("Interrupted");
+				Thread.currentThread().interrupt();
```
eaglerainbow: see above

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
```diff
@@ -308,6 +310,7 @@ public void testLockObjectCompositeWorks() throws InterruptedException {
...
```
eaglerainbow: see above

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
```diff
@@ -423,7 +423,7 @@ private boolean isApplicationInScrapableState(String state) {
 		}
 		
 		/* TODO: To be enhanced, once we know of further states, which are
-		 * also scrapable.
+		 * also scrapeable.
```
eaglerainbow: a nice one :) 😁

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
warlockza: Should we just take out the assert.fail and interrupt?

warlockza: Have put the Interrupt before the Assert, that way , we are still being good Thread Citizens by not dropping the interrupted status, and still allowing our assert to occur.

eaglerainbow: That's a good compromise

src/test/java/org/cloudfoundry/promregator/cache/AutoRefreshingCacheMapTest.java
```diff
@@ -132,7 +132,8 @@ public void run() {
 				value = subject.get("global");
 				Assert.assertEquals("labolg", value);
 			} catch (InterruptedException e) {
-				Assert.fail("Thread was interrupted");
+				Thread.currentThread().interrupt();
+                Assert.fail("Thread was interrupted");
```
eaglerainbow: Please check indention.

warlockza: Fixed now - Please confirm

warlockza: Can this be merged?

PR #132, opened 2019-11-30
Fixed squid:S1155 and removed some throws that are not needed

warlockza:
Fixed squid:S1155
Removed some throws that are not needed
Fixed @deprecated javadocs
Removed some unneeded imports (that I forgot about in previous pulls)
Added private constructors to static utility classes
Removed a variable that was being set but never used

PR #137, opened 2019-12-01
Properly commenting further parts for unsupported timestamp

eaglerainbow:
## Current Situation
#132 [removed a line](https://github.com/promregator/promregator/pull/132/files#r352376538), which isn't used today, but eventually would, as soon as simpleclient supports (optional) timestamps.

## Problem Statement
Information about the right implementation partially gets lost.

## Solution Approach
Add the coding again, but comment it to ensure that it does not trigger SonarQube messages.

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/fetcher/TextFormat004Parser.java
+++ b/src/main/java/org/cloudfoundry/promregator/fetcher/TextFormat004Parser.java
@@ -109,11 +109,11 @@ public class TextFormat004Parser {
 		double value = 0.0f;
 		String valueString = null;
 		Matcher mValueText = PATTERN_PARSE_VALUETEXT.matcher(rest);
-		int end = 0;
+		// int end = 0;
 		if (mValueText.find()) {
 			valueString = mValueText.group(1);
 			value = this.parseGoDouble(valueString); // NB: an exception cannot be thrown here (and the exception in fact is an Error)
-			end = mValueText.end();
+			// end = mValueText.end();
 		} else {
 			Matcher mValue = PATTERN_PARSE_VALUE.matcher(rest);
 			if (!mValue.find()) {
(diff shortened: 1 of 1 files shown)


eaglerainbow:
@warlockza I'd be interested in getting your feedback here.

warlockza:
When you say Java Simpleclient? What component specifically are you referring to...

Is there sanity in having an interface, with having one parser for the existing approach and one for the upcoming approach? That way we are not commenting out unused code, not upsetting sonar, not removing the code (and potentially losing it ) and have two complete side,by-side implementations (one known to be currently working, and one that could be substituted in , in the future or even made configurable)

eaglerainbow:
> When you say Java Simpleclient? What component specifically are you referring to...

I refer to Prometheus Java Client library called "simpleclient"  - see also https://github.com/prometheus/client_java
Promregator uses this library to both expose its own metrics. It also uses it for handling metrics values (and its sets) internally. 

> Is there sanity in having an interface, with having one parser for the existing approach and one for the upcoming approach?

Well, life is a little more complex than this. There is an [official standard defined on the protocol](https://github.com/prometheus/docs/blob/1ae8cfae5eeff995d3fe7178b382b105db0704ec/content/docs/instrumenting/exposition_formats.md) that Prometheus speaks. That standard defines that there is an optional field for the timestamp. However, the simpleclient (see above) does not support timestamps. That's it does not make sense to extract the information (as nobody would be there within Promregator that is capable to handle it internally). That's why it is (so far) still commented - we would not know where to write the data (an hence it would be dead code, which is even executed).
In short: Promregator's text-004 parser is better than the one provided by Prometheus on the platform. 

Hiding this behind an own interface looks rather over-engineered to me: As it is an optional field, it can be handled in a compatible manner.

PR #141, opened 2020-01-03
Correct metrics_path back to original

checketts:
Prometheus discovery was broken by a previous cleanup. This restores it to the previous behavior

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/endpoint/DiscoveryEndpoint.java
+++ b/src/main/java/org/cloudfoundry/promregator/endpoint/DiscoveryEndpoint.java
@@ -91,7 +91,7 @@ public class DiscoveryEndpoint {
 			return __meta_promregator_target_instanceId;
 		}
 		
-		public String get__metrics_path() {
+		public String get__metrics_path__() {
 			return this.__meta_promregator_target_path;
 		}
 	}


eaglerainbow:
846167f2885343927e7d4d0bda8c52daddc212a6 is available on `master` only so far. 
No release version was built => only affecting mainline but no officially released version yet.

eaglerainbow:
Triggered voter manually

checketts:
I would be interested in adding some endpoint/mockmvc tests around this behavior. Would that be welcome? I see there isn't any current MockMVC tests in the project.

eaglerainbow:
@checketts Be my guest on this! 
Perhaps you may also want to check #142, which I just raised out of this incident here.

checketts:
Sounds good! Would you be opposed to me upgrading to Springboot 2.2, JUnit 5, and using Kotlin for those tests?

eaglerainbow:
Spring boot 2.2 and Junit 5 is no problem. We haven't used Kotlin here (and I also would require education on this), so the latter could be a challenge...

eaglerainbow:
Be warned for the attempt to bump spring-boot  to 2.2. We have spring-cloud in place and I had gained bad experience with it (see also 90fe2ff477e85449cf27b06857d8280d5150f091 )

checketts:
I'll try the upgrades in one PR, and do a follow up with the Kotlin tests.

checketts:
I'm not clear on the 'missing adoption' issue. Can you add some details?

eaglerainbow:
> I'm not clear on the 'missing adoption' issue with Spring Cloud.

In a nutshell:
* When specifying dependency to spring-cloud, you must explicitly mention which version you want to use. There is no dependencyManagement provided by spring-boot.
* Promregator uses spring-cloud for some special case, so we have such a dependency.
* Due to incompatible changes between the spring-boot versions, there is always only one unique Spring-Cloud version which works with that spring-boot version.
* The version naming schema of spring cloud is different than for spring-boot.
* Even worse. spring-boot 2.1.11 apparently changes something on which spring-cloud `Greenwich.SR*` is relying on. 

==> we are somehow trapped there.

I haven't checked on the situation with spring-boot:2.2 and spring-cloud:??? yet.

checketts:
Challenge accepted ;)

eaglerainbow:
spring-cloud's release train for spring-boot:2.2 is called Hoxton - source: https://spring.io/projects/spring-cloud

eaglerainbow:
Please feel free to continue our discussion at #142 
We can also raise an own issue for bumping to spring-boot 2.2 (et al).

PR #143, opened 2020-01-03
Add config ability to disable single endpoint

checketts:
## Current Situation
When discovering #141 , it caused all my targets to fallback to using the Single Endpoint Scraping which caused very heavy load on my Promregator instances.

## Solution Approach
Since I don't want any of my Prometheus instances or users to use Single Endpoint Scraping, I would like a way to disable it.

This approach uses a max value, so a user could set a limit and if they cross it (because more and more apps have been added to cloud foundry), they would be notified by failures instead of the entire service suffering.

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/endpoint/MetricsEndpoint.java
+++ b/src/main/java/org/cloudfoundry/promregator/endpoint/MetricsEndpoint.java
@@ -6,6 +6,7 @@ import java.util.List;
 import org.apache.log4j.Logger;
 import org.cloudfoundry.promregator.fetcher.MetricsFetcher;
 import org.cloudfoundry.promregator.scanner.Instance;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Scope;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
@@ -28,6 +29,9 @@ import io.prometheus.client.exporter.common.TextFormat;
 @RequestMapping(EndpointConstants.ENDPOINT_PATH_SINGLE_ENDPOINT_SCRAPING)
 public class MetricsEndpoint extends AbstractMetricsEndpoint {
 	private static final Logger log = Logger.getLogger(MetricsEndpoint.class);
+
+	@Value("${promregator.single.endpoint.max:-1}")
+	private int maxAllowedForSingleEndpoint;
 
 	@GetMapping(produces=TextFormat.CONTENT_TYPE_004)
 	public ResponseEntity<String> getMetrics() {
@@ -63,6 +67,10 @@ public class MetricsEndpoint extends AbstractMetricsEndpoint {
 	 */
 	@Override
 	protected List<MetricsFetcher> createMetricsFetchers(List<Instance> instanceList) {
+		if(maxAllowedForSingleEndpoint != -1 && instanceList.size() > maxAllowedForSingleEndpoint) {
+			throw new RuntimeException("promregator.single.endpoint.max is configured to only allow "+maxAllowedForSingleEndpoint+" instances but there were "+ instanceList.size());
+		}
+
 		if (instanceList.size() > 20) {
 			log.warn(String.format("You are using Single Endpoint Scraping with %d (>20) active targets; to improve scalability it is recommended to switch to Single Target Scraping", instanceList.size()));
 		}


src/main/java/org/cloudfoundry/promregator/endpoint/MetricsEndpoint.java
```diff
@@ -63,6 +67,10 @@ protected boolean isLabelEnrichmentSuppressable() {
 	 */
 	@Override
 	protected List<MetricsFetcher> createMetricsFetchers(List<Instance> instanceList) {
+		if(maxAllowedForSingleEndpoint != -1 && instanceList.size() > maxAllowedForSingleEndpoint) {
+			throw new RuntimeException("promregator.single.endpoint.max is configured to only allow "+maxAllowedForSingleEndpoint+" instances but there were "+ instanceList.size());
```
eaglerainbow: Better use `String.format` here.

eaglerainbow: This will cause a SonarQube finding. Creating a dedicated exception for this is recommended then (may be private and local to the class here).

src/main/java/org/cloudfoundry/promregator/endpoint/MetricsEndpoint.java
```diff
@@ -29,6 +30,9 @@
 public class MetricsEndpoint extends AbstractMetricsEndpoint {
 	private static final Logger log = Logger.getLogger(MetricsEndpoint.class);
 
+	@Value("${promregator.single.endpoint.max:-1}")
```
eaglerainbow: I am not convinced about the name yet. what does "maximal endpoint" mean? How many endpoints we have? No, apparently this not meant here. 
Perhaps `promregator.endpoint.single.maxTargets`?

eaglerainbow: If we introduce a new config option, then it also has to be explained in https://github.com/promregator/promregator/blob/3330cd742f3710b89a855cdf80c47f2e2f116f0c/docs/config.md

src/main/java/org/cloudfoundry/promregator/endpoint/MetricsEndpoint.java
```diff
@@ -29,6 +30,9 @@
...
+	private int maxAllowedForSingleEndpoint;
```
eaglerainbow:
```suggestion
	private int maxTargetsAllowed;
```

src/main/java/org/cloudfoundry/promregator/endpoint/MetricsEndpoint.java
eaglerainbow: I am asking myself right now whether it wouldn't be better to send a proper error message with a good HTTP status code here instead.

checketts:
Let me explain better: In my setup I have a cron that hits my various Promregators' discovery endpoints and saves the json local to Prometheus. Then in the Prometheus config I have something like:

```
scrape_configs:
  - job_name: "north"
    scheme: https
    file_sd_configs:
      - files:
        - "north.json"
    relabel_configs:
     - source_labels: [__meta_promregator_target_applicationName,__meta_promregator_target_instanceNumber]
       target_label: instance
       separator: "-"
     - source_labels: [__meta_promregator_target_orgName]
       target_label: org_name
     - source_labels: [__meta_promregator_target_spaceName]
       target_label: space_name
       target_label: application
```

Since I hadn't set a `metrics_path`, it had fallen back to that URL (`/metrics`) when the `__metrics_path__` label stopped matching. That is the default `metrics_path` since I hadn't set it anywhere myself. So that unintentionally sent 60 targets calling single endpoint for scraping.

Did that help clarify?

src/main/java/org/cloudfoundry/promregator/endpoint/MetricsEndpoint.java
checketts: It would be. A `503` would probably be a good idea. If you think this feature is worth having.

checketts: Sorry I'm not sure I understand the reasoning. Is `String.format` preferred for performance reasons or for i18n?

I'm happy to oblige. I just want to understand why it is preferred.

eaglerainbow:
> Did that help clarify?

Yes, it greatly does. 

Mmmh... the right solution would be to have the SingleMetricsEndpoint registered to something non-standard as `/metrics` -- to prevent that this is happening again. However, that would be hell-of-incompatible for the current consumers => not an option.

eaglerainbow:
I am currently still struggling with this configuration option: The default is de facto "infinite". So, the "emergency break" only would help, if users configured it during at configuration time. Given the perceived-hundreds of configuration options Promregator has, I doubt that they will find that switch - and if they do, they'd be confused why you should limit the number of targets to be scraped (they expect that the number will increase over time for sure). I fear that they will either put it to some ridiculously high value (say 1000) "to be safe", or they will not set it at all.
Moreover, they will have a hard time to understand why they need to configure this "safety belt" for the Single Scraping Endpoint, but not for the Single Target Endpoint. Only after the intensive study of the difference between these two methods, they might come to the point.

eaglerainbow:
I am currently more into another idea: Why not switching off `/metrics` completely? If you are on Single Target Scraping, then it's a good safety measure. Moreover, it's easy to understand and we can provide simple guidance in documentation that. 
If the situation you have encountered happened again, this approach would also have stopped you from killing Promregator. We could raise a lovely "404" plus some JSON-formatted error text indicating that due to the configuration option "xyz" this endpoint is disabled.

checketts:
I'm good with entirely disabling the endpoint. I'll update the PR to support that direction.

src/main/java/org/cloudfoundry/promregator/endpoint/MetricsEndpoint.java
eaglerainbow: It's a kind of that 😄 

`String.format` is indeed supportive on i18n. However, Promregator is not i18n-enabled yet. I also don't expect to see happening that soon. Yet, I don't want to rule it out completely. If we then have many string concatenations all around, this wouldn't make our life easier. 

Moreover, I have seen too many cases (in my business life) where people started to "glue strings together", which they then spooled out as human readable text. By this they "hard-code the English grammar" (it's even worse when they do so writing "source code in German"). That is why I defined a rule on this to myself (which helped me getting along well so far): "If you have a placeholder in your text, then always use `String.format` (at least)". So, you might call it "experience out of pain" 😀 - or to nip things in the bud.

Is that reasoning okay with you?

checketts: Certainly. The reason I asked is with slf4j and log4j2 you can do the following with logging: 

```
log.debug("Creating Metrics Fetcher for instance {}", instance.getInstanceId());
```

And I was considering making the recommendation eventually. (Obviously we would still stick to `String.format` for the exception messages though)

eaglerainbow: Yeah, I know: for the logging facilities you have this alternative - which I would not disregard in general. However, to be honest, I am not used to this new feature - and I also had some bad experience (on serializing these kind of log records) in one project with it. That's why I am not using it so far (once bitten, twice shy).

checketts:
Closing this since we are planning on removing the endpoint completely

PR #144, opened 2020-01-03
Upgrade SpringBoot to 2.2.2.RELEASE and Spring Cloud to Hoxton

checketts:
Upgrade went smoothly. Tests passed locally and appear to be running fine. 

Let me know if you would prefer the JUnit upgrade in this PR, for now I'm planning on keeping it separate. There will be a lot more changes there since the tests will need tweaking.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,8 +38,7 @@
 	<parent>
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
-		<version>2.1.10.RELEASE</version> <!-- see also https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-parent -->
-		<!-- Warning! Cannot bump to 2.1.11.RELEASE due to missing adoption by spring-cloud.version Greenwich.SR4 -->
+		<version>2.2.2.RELEASE</version> <!-- see also https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-parent -->
 		<relativePath /> <!-- lookup parent from repository -->
 	</parent>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -38,16 +38,15 @@
 	<parent>
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
-		<version>2.1.10.RELEASE</version> <!-- see also https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-parent -->
-		<!-- Warning! Cannot bump to 2.1.11.RELEASE due to missing adoption by spring-cloud.version Greenwich.SR4 -->
+		<version>2.2.2.RELEASE</version> <!-- see also https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-parent -->
```
eaglerainbow: We can get rid of the line at https://github.com/promregator/promregator/pull/144/files#diff-600376dffeb79835ede4a0b285078036R53

pom.xml
```diff
@@ -38,16 +38,15 @@
...
 		<relativePath /> <!-- lookup parent from repository -->
 	</parent>
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 		<java.version>1.8</java.version>
-		<spring-cloud.version>Greenwich.SR4</spring-cloud.version>
+		<spring-cloud.version>Hoxton.RELEASE</spring-cloud.version>
```
eaglerainbow: Is there a special reason why you don't like `Hoxton.SR1`?
See https://mvnrepository.com/artifact/org.springframework.cloud/spring-cloud-dependencies/Hoxton.SR1

pom.xml
eaglerainbow: The version of awaitility is reported to be superfluous now (see also https://github.com/promregator/promregator/pull/144/files#diff-600376dffeb79835ede4a0b285078036R201 )

eaglerainbow:
> Let me know if you would prefer the JUnit upgrade in this PR, for now I'm planning on keeping it separate. There will be a lot more changes there since the tests will need tweaking.

I am fine to have it separate, esp. if you expect many changes to become necessary.

pom.xml
checketts: Nope. I had overlooked that release number. I'll switch to that

eaglerainbow:
Let's see what the voter says....

PR #145, opened 2020-01-07
Sonar Cleanups (Modifiers in interfaces, blocks for simple If statements, Null => Empty Collections)

warlockza:
Removed unneeded modifiers on interfaces
Put Braces around Simple if statements
Corrected the returning of Null collections to instead return empty collections

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/config/Target.java
+++ b/src/main/java/org/cloudfoundry/promregator/config/Target.java
@@ -7,6 +7,8 @@ import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 
 import org.apache.log4j.Logger;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 public class Target {
 	private static final Logger log = Logger.getLogger(Target.class);
@@ -143,8 +145,9 @@ public class Target {
 
 	/**
 	 * @return the preferredRouteRegex
+	 * This will never return a null value
 	 */
-	public List<String> getPreferredRouteRegex() {
+	public @NonNull List<String> getPreferredRouteRegex() {
 		if (this.preferredRouteRegex == null) {
 			return Collections.emptyList();
 		}
@@ -160,7 +163,11 @@ public class Target {
 		this.cachedPreferredRouteRegexPattern = null; // reset cache
 	}
 
-	public List<Pattern> getPreferredRouteRegexPatterns() {
+	/**
+	 * @return the list of preferred Route Regex Patterns
+	  * This will never return a null value
+	 */
+	public @NonNull List<Pattern> getPreferredRouteRegexPatterns() {
 		if (this.cachedPreferredRouteRegexPattern != null) {
 			return this.cachedPreferredRouteRegexPattern;
 		}
--- a/src/main/java/org/cloudfoundry/promregator/config/Target.java
+++ b/src/main/java/org/cloudfoundry/promregator/config/Target.java
@@ -1,6 +1,7 @@
 package org.cloudfoundry.promregator.config;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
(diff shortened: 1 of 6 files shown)


src/main/java/org/cloudfoundry/promregator/config/Target.java
```diff
@@ -145,7 +146,7 @@ public void setAuthenticatorId(String authenticatorId) {
 	 */
 	public List<String> getPreferredRouteRegex() {
 		if (this.preferredRouteRegex == null) {
-			return null;
+			return Collections.emptyList();
```
eaglerainbow: Nope. That's a change of semantics: An empty list of PreferredRoutRegex is something different than none of them (`null`). See also https://github.com/promregator/promregator/blob/94688a4530a6d58d12e4adf74c21ab68c3b31a82/src/main/java/org/cloudfoundry/promregator/config/Target.java#L168

PS: I thought that @checketts already had had a similar suggestion before, which I rejected. However, I couldn't find it anymore.

src/main/java/org/cloudfoundry/promregator/endpoint/EndpointConstants.java
```diff
@@ -1,11 +1,11 @@
 package org.cloudfoundry.promregator.endpoint;
 
 public interface EndpointConstants {
-	static final String ENDPOINT_PATH_DISCOVERY = "/discovery";
-	static final String ENDPOINT_PATH_SINGLE_ENDPOINT_SCRAPING = "/metrics";
-	static final String ENDPOINT_PATH_SINGLE_TARGET_SCRAPING = "/singleTargetMetrics";
-	static final String ENDPOINT_PATH_PROMREGATOR_METRICS = "/promregatorMetrics";
-	static final String ENDPOINT_PATH_CACHE_INVALIDATION = "/cache/invalidate";
+	String ENDPOINT_PATH_DISCOVERY = "/discovery";
```
eaglerainbow: I still don't get it that attributes of interfaces in Java are implicitly `final`, but be it...

src/main/java/org/cloudfoundry/promregator/config/Target.java
checketts: Nope. I don't recall making any proposal around that.

eaglerainbow: Strange 😕 thought so... but doesn't matter... Thanks for the pointer, though!

warlockza: I understand that it is a change of semantics, however I traced it through the codepath as to where it would be used, in ReactiveAppInstanceScanner.javam line +/- 322

`if (patterns == null || patterns.isEmpty()) {
			log.debug("No Preferred Route URL (Regex) provided; taking first Application Route in the list provided");
			return urls.get(0);
		}`

And the code treats empty list and null the same..

eaglerainbow: I get your point. However, then we should make this consistent also into the other direction:
* Declare the return of `getPreferredRouteRegexPatterns` and `getPreferredRouteRegex` to be `@NonNull`
* Document that briefly in a JavaDoc comment.
* Remove the `patterns == null` comparison at https://github.com/promregator/promregator/blob/58a69f9ecd1f0a15310086a0d3bc5ed6cdc54a0e/src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java#L322  as they are redundant.
* If we find further spots like this (where we are checking something on `null` which can't be `null`, we should get rid of them as well.

What do you think?

src/main/java/org/cloudfoundry/promregator/endpoint/EndpointConstants.java
warlockza: I found this handy description for interfaces

All the members (methods and fields) of an interface are public.
All the methods in an interface are public and abstract (except static and default).
All the fields of an interface are public, static and, final by default.

src/main/java/org/cloudfoundry/promregator/config/Target.java
warlockza: I wholeheartedly concur

eaglerainbow:
@checketts @warlockza 
I am currently playing around a little with Discord (an own test server at https://discord.gg/7ugjZh) for Promregator. Feel free to join me there, if you are interested...

PR #146, opened 2020-01-07
Initial pass of updating tests to run with JUnit Jupiter

checketts:
The build is already setup to leverage JUnit 5. So merely having to update tests to leverage the runner and improve the assertions. 

I didn't finish all the tests yet. But this is a good state that can be merged. So I can branch off it and begin the MockMVC based tests.


src/test/java/org/cloudfoundry/promregator/MonoErrorHandlingTest.java
```diff
@@ -3,16 +3,16 @@
...
 import reactor.core.publisher.Mono;
 
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
 public class MonoErrorHandlingTest {
 	@Test
 	public void test() {
-		boolean raised = false;
-		try {
+		assertThatThrownBy(() ->{
```
checketts: Switched to the `assertThatThrownBy` mechanism instead of our own try/catch

src/test/java/org/cloudfoundry/promregator/SpringBootLoadPropertiesForTesting.java
```diff
@@ -1,56 +1,50 @@
...
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.not;
 
 @SpringBootTest(classes = { SpringBootLoadPropertiesForTestingSpringApplication.class })
-@RunWith(SpringRunner.class)
+@ExtendWith(SpringExtension.class)
 @ContextConfiguration(initializers = { ConfigFileApplicationContextInitializer.class })
 @ActiveProfiles(profiles= {"springBootLoadPropertiesForTesting"})
+@TestPropertySource(properties = "encrypt.key=mySecretKey")
```
checketts: Eliminated the `EnvironmentVariables` approach, to just using Spring properties directly.

src/test/java/org/cloudfoundry/promregator/auth/AuthenticatorControllerTest.java
```diff
@@ -25,33 +27,33 @@
...
-		Assert.assertNotNull(subject);
+		assertThat(subject).isNotNull();
 		
 		AuthenticationEnricher auth0 = subject.getAuthenticationEnricherById("unittestAuth0");
-		Assert.assertTrue(auth0 instanceof BasicAuthenticationEnricher);
+		assertThat(auth0).isInstanceOf(BasicAuthenticationEnricher.class);
 		
 		AuthenticationEnricher auth1 = subject.getAuthenticationEnricherById("unittestAuth1");
-		Assert.assertTrue(auth1 instanceof OAuth2XSUAAEnricher);
+		assertThat(auth1).isInstanceOf(OAuth2XSUAAEnricher.class);
```
checketts: Improved the assertions so if the boolean check fails, the logs show the state of the variables instead of just a 'expected true was false' message

src/test/java/org/cloudfoundry/promregator/auth/OAuth2XSUAAEnricherTest.java
```diff
@@ -56,25 +58,11 @@ public void testAppropriateJWTCall() {
...
-		ArgumentCaptor<String> keyCaptor = ArgumentCaptor.forClass(String.class);
-		ArgumentCaptor<String> valueCaptor = ArgumentCaptor.forClass(String.class);
-		Mockito.verify(mockGet).setHeader(keyCaptor.capture(), valueCaptor.capture());
-		String value = valueCaptor.getValue();
-		Assert.assertEquals("Bearer someAccessToken", value);
-		
-		String key = keyCaptor.getValue();
-		Assert.assertEquals("Authorization", key);
+
+		verify(mockGet).setHeader("Authorization", "Bearer someAccessToken");
```
checketts: This single verification line does the same as the captors/verification

src/test/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeineTimeoutSpringApplication.java
```diff
@@ -20,21 +20,9 @@ public InternalMetrics internalMetrics() {
...
-		}
-	}
-	
 	@Bean
 	public CFAccessor parentMock() {
-		CFAccessor mock = Mockito.mock(CFAccessor.class);
-		Mockito.when(mock.retrieveOrgId("dummy")).then(new TimeoutMonoAnswer());
-		Mockito.when(mock.retrieveSpaceId("dummy1", "dummy2")).then(new TimeoutMonoAnswer());
-		Mockito.when(mock.retrieveAllApplicationIdsInSpace("dummy1", "dummy2")).then(new TimeoutMonoAnswer());
-		Mockito.when(mock.retrieveSpaceSummary("dummy")).then(new TimeoutMonoAnswer());
```
checketts: Moved the mocking calls into the `@BeforeEach` method of the test so verifications are reset properly between tests.

src/test/java/org/cloudfoundry/promregator/MonoErrorHandlingTest.java
```diff
@@ -3,16 +3,16 @@
...
 
-import org.junit.Assert;
-import org.junit.Test;
 
+import org.junit.jupiter.api.Test;
 import reactor.core.publisher.Mono;
 
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
 public class MonoErrorHandlingTest {
```
eaglerainbow: minor: 😀 This class was once a playground for Mono's, because I did not trust the reactorProject properly... 
Perhaps we can even think about getting rid of it completely...

src/test/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeineTimeoutTest.java
```diff
@@ -31,15 +36,29 @@
 	
 	@Autowired
 	private CFAccessorCacheCaffeine subject;
-	
-	@Before
-	public void invalidateCaches() {
+
+	@BeforeEach
+	public void setup() {
+		Mockito.reset(this.parentMock);
```
eaglerainbow: Just a stupid idea: The cache stuff below has nothing to do with this resetting/configuring of the mock. Hence, what about the idea of having two separate `@BeforeEach` methods for that to properly separate the semantics?

src/test/java/org/cloudfoundry/promregator/SpringBootLoadPropertiesForTesting.java
eaglerainbow: Thanks for the explanation. I am not 100% convinced yet that this will work out properly. I will check further on this to see if I am right or wrong and will let you know the result.

src/test/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeineTimeoutTest.java
checketts: I'm fine with that. I renamed the method to `setup` since it was mixing concerns. I'll actually be switching to `@MockBean` in an upcoming PR so the reset will probably be removed at that time.

src/test/java/org/cloudfoundry/promregator/MonoErrorHandlingTest.java
checketts: I've deleted it now

src/test/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeineTimeoutTest.java
checketts: Done

src/test/java/org/cloudfoundry/promregator/SpringBootLoadPropertiesForTesting.java
eaglerainbow: Luckily that works correctly.

src/test/java/org/cloudfoundry/promregator/SpringBootLoadPropertiesForTesting.java
```diff
@@ -1,56 +1,50 @@
...
 
 	@Test
 	public void testContextLoadsWithEncryptedValue() {
 		String secretValue = springBootLoadPropertiesForTestingSpringApplication.getSecretValue();
-		assertThat("passwords do not match but should", secretValue, is("mysecret"));
+		assertThat(secretValue).as("passwords do not match but should").isEqualTo("mysecret");
 	}
 
 	@Test
 	public void testContextLoadsWithEncryptedValueNotMatching() {
```
eaglerainbow: Note to myself: The more I think about this test, the less I think it's worthwhile having it.

eaglerainbow: Fixed with c7e301932f9109ab516a10aa796506e84c7a2a03

Issue #147, opened 2020-01-08
org.cloudfoundry.promregator.PromregatorApplicationSimulatorTest blocks mainline


eaglerainbow:
## Environment
Promregator@master

## Summary / Problem Statement
Unit Test in main line currently blocked

## Observed Behavior
```
22:07:52 [ERROR] Errors: 
22:07:52 [ERROR]   PromregatorApplicationSimulatorTest.contextLoads ? IllegalState Failed to load...
22:07:52 [ERROR]   PromregatorApplicationSimulatorTest.testDiscoveryWorks ? IllegalState Failed t...
22:07:52 [ERROR]   PromregatorApplicationSimulatorTest.testSingleInstance ? IllegalState Failed t...
22:07:52 [ERROR]   PromregatorApplicationTest.contextLoads ? IllegalState Failed to load Applicat...
```

with 
```
java.lang.IllegalStateException: Failed to load ApplicationContext
	at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContext(DefaultCacheAwareContextLoaderDelegate.java:132)
	at org.springframework.test.context.support.DefaultTestContext.getApplicationContext(DefaultTestContext.java:123)
	at org.springframework.test.context.web.ServletTestExecutionListener.setUpRequestContextIfNecessary(ServletTestExecutionListener.java:190)
	at org.springframework.test.context.web.ServletTestExecutionListener.prepareTestInstance(ServletTestExecutionListener.java:132)
	at org.springframework.test.context.TestContextManager.prepareTestInstance(TestContextManager.java:244)
	at org.springframework.test.context.junit.jupiter.SpringExtension.postProcessTestInstance(SpringExtension.java:98)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$invokeTestInstancePostProcessors$5(ClassBasedTestDescriptor.java:337)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.executeAndMaskThrowable(ClassBasedTestDescriptor.java:342)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$invokeTestInstancePostProcessors$6(ClassBasedTestDescriptor.java:337)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
[...]

PR #148, opened 2020-01-08
Change internal logger to slf4j since we are using log4j-over-slf4j already

checketts:
This just updates the logger to remove confusion and be more efficient. 

I didn't remove the log4j-over-slf4j dependency since I wasn't certain if some log4j loggers were in use by third party libs.

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
+++ b/src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
@@ -8,7 +8,6 @@ import java.util.concurrent.Executors;
 
 import javax.annotation.PostConstruct;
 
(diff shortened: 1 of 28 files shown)


eaglerainbow:
> I didn't remove the log4j-over-slf4j dependency since I wasn't certain if some log4j loggers were in use by third party libs.

If you have found all occurrences, then I think you should be brave here. We have a rather consistent BOM here and if we had one library, which still requires log4j, then I want to know it asap.

eaglerainbow:
Voter is red - checking on the cause...

eaglerainbow:
looks very much like #147 
I recommend merging/rebasing this PR onto the current master.

PR #156, opened 2020-02-20
Adding Kotlin Infrastructure / Adding initial MockMVC test for Discovery endpoint

checketts:
## Current Situation
Add MockMVC based endpoint tests

## Problem Statement
A refactor broke the discovery endpoint briefly on master. Adding MockMVC based endpoint tests will ensure that doesn't happen again. It also gives us a way to document the behavior of the various endpoints and test the application without having to start the complete app.

## Solution Approach
Add MockMVC unit tests

## Prerequisites
Not required, but this PR also introduces Kotlin so the project has a place to evaluate it and determine if more of the codebase could be converted to use it.


eaglerainbow:
Thanks for the proposal! However, I currently have to admit that my kotlin skills/experience is way too little to check for any issues. I will have to learn some stuff first, before I can really judge that.
However, any kotlin-fan (also outside of this project) may perform a review here as well and provide feedback!

src/test/java/org/cloudfoundry/promregator/endpoint/DiscoveryEndpointControllerTest.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest
+import org.springframework.test.context.junit.jupiter.SpringExtension
+import org.springframework.test.web.servlet.MockMvc
+import org.springframework.test.web.servlet.get
+
+@ExtendWith(SpringExtension::class)
+@WebMvcTest(DiscoveryEndpoint::class, properties = ["promregator.simulation.enabled=true", "promregator.lifecycle.enabled=false"])
+class DiscoveryEndpointControllerTest(
+       @Autowired private val mockMvc: MockMvc
```
checketts: Here is a Kotlin overview to get you up to speed (and to highlight some of the benefits):

1 - You can define properties in your constructor (`private val` is essentially creating a field in the test for `mockMvc`, and flagging it as final so it can't be changed. It also creates getters/setters that Java code could call, if it weren't `private`. The `val`/`var` support in Kotlin allows the compiler to do additional checks against type and nullability as well (aka 'smart casts').

2 - Constructor injection is supported in JUnit 5 tests (which lets Spring dependency injection work, and still make them `val`s

src/test/java/org/cloudfoundry/promregator/endpoint/DiscoveryEndpointControllerTest.kt
```diff
@@ -0,0 +1,65 @@
...
+) {
+    @MockkBean private lateinit var discoverer: CFMultiDiscoverer
+    @MockkBean(relaxed = true) private lateinit var config: PromregatorConfiguration
```
checketts: `@MockkBean` is equivalent to Creating a `@Bean` that is a mock, and is being reset after each test. This technique saves us from creating extra `@Configurations` for beans that aren't used elsewhere.

src/test/java/org/cloudfoundry/promregator/endpoint/DiscoveryEndpointControllerTest.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    @BeforeEach
+    fun setup() {
+
+    }
+
+    /** Test for [DiscoveryEndpoint.getDiscovery] */
+    @Test
+    fun `get returns 503 when no targets are resolvable`() {
+        every {discoverer.discover(any(), any()) } returns listOf()
```
checketts: Since Kotlin reifies types (addressing type erasure that occurs with generics), it can support simplified mocking like these `any()` calls here. Similar Java mocking would have requires `anyString()`/`anyInt()`/`any<MyObject>()` and other similar approaches. 

When a lambda is the last argument of a method call, Kotlin let's up leave the parenthesis off, or more the lambda outside parens.

The following 3 are identical, but allow for leaving off noise
```
every({ discoverer.thing() }) returns "thing"
every { discoverer.thing() } returns "thing"
every() { discoverer.thing() } returns "thing"
```

src/test/java/org/cloudfoundry/promregator/endpoint/DiscoveryEndpointControllerTest.kt
```diff
@@ -0,0 +1,65 @@
...
+
+        mockMvc.get("/discovery").andDo { print() }.andExpect {
+            status { isServiceUnavailable }
+        }
+    }
+
+    /** Test for [DiscoveryEndpoint.getDiscovery] */
+    @Test
+    fun `get returns json that includes the apps space, instance, and metrics path`() {
```
checketts: Kotlin functions support spaces when using backticks. Not useful in app code, but is quite useful in tests

eaglerainbow:
Thanks for your effort and the insights! 
I am currently running myself through the kotlin language and some tutorials (where I find it surprising that for the latter I did not find anything consistent&understandable so far that a Java developer like me simply could run through without bumping into words which have not been defined so far -- where is our mathematical heritage gone to 😖?). However, the myriad of special keywords and concepts (e.g. `companion`, `infix`, `operator`, `lateinit`, `lazy`, `get()`/`set()`-generation and declaration, sealed, extension, implicitly boxed, no automatic cast extension, named parameters, in fact object orientation, but notation in functional style, ...) causes me headache. 
Therefore, I'll need some more time to learn&digest the language and then will come back to you...

checketts:
Wow you are diving in headfirst to begin running into `infix`, `operator`, and `sealed` already. I tried to keep this PR to intro concepts. ;)

Here is a great intro video by Venkat Subramanian: https://www.youtube.com/watch?v=fsNhjYgXxlg of Kotlin for Java Developers.

I apologize, I should have included a short explanation of 'why' Kotlin would be beneficial:

* Nullability is built into the type system (`String` is different from `String?` (the nullable type))
* Data classes make for succinct POJO declarations (particularly in the `ConfigurationProperties`)
* Named/Default parameters make object creation (particularly in tests) less noisy

There are many more reasons (improved mocking, extension methods, improved collections apis), but I don't want to overwhelm you! :D Also several like `companion`, `operator`, `infix` aren't really needed in Promregator for now.

eaglerainbow:
> Wow you are diving in headfirst to begin running into infix, operator, and sealed already. I tried to keep this PR to intro concepts. 😉

I saw that 😁  But you are also changing our build infrastructure, and I know what you still have in your pipeline. And foremost, I want you to get a proper code review here. So, the only chance is "up-skilling"... 😀 
In fact, this PR is about whether in Promregator we want to go the way of Kotlin, or whether we should keep plain-Java. So, for that decision we need a holistic view on it.

> Venkat Subramanian: https://www.youtube.com/watch?v=fsNhjYgXxlg

I am just watching it; though he's much better than most of the rest, he still has the same issue , for example at time index 37:50 (talking about `until` as infix while he's explaining how `for` loops work in kotlin 😠 )
I don't know whether I'll be able to listen to him within one shot... (time on Promregator is a little restricted for me right now - has multiple smaller reasons).

> Nullability is built into the type system

In general, I agree to this as a "Good Thing™️". Note though that in Promregator with currently have little problems with that (from point of view of our consumers). Making our life easier by having the compiler checked if an NPE could occur, is valuable, though.

> Data classes make for succinct POJO declarations

Indeed, this could be helpful. Also the `Target` and `ResolvedTarget` could benefit from that.

> Named/Default parameters make object creation (particularly in tests) less noisy

Being a Kotlin beginner up to now, I am not 100% sure if I will like this feature. I already know the concept of defaulted and named parameters from ABAP (in fact naming parameters is enforced there, sometimes making refactoring hell). My experience is very mixed there; it tends to let foolish developers create prototypes (of functions/methods), which only looks simple but in fact is flooded with complexity. Sometimes it is also abused for retaining compatibility where a proper refactoring would have been due. 
If you don't know ABAP, you may see similar issues also (for instance) in python with APIs in the numpy package.

> Also several like companion, operator, infix aren't really needed in Promregator for now.

I tend to agree on this as well. Yet, I think I need to understand the power of the concepts at least from a perspective of a "general overview" to be able to judge the direction properly.

pom.xml
```diff
@@ -56,6 +56,7 @@
 		<prometheus.version>0.6.0</prometheus.version>
 		<cloudfoundry.client.version>3.21.0.RELEASE</cloudfoundry.client.version>
 		<logback.contrib.version>0.1.5</logback.contrib.version>
+		<kotlin.version>1.3.61</kotlin.version>
```
eaglerainbow: If we add kotlin, we also should add support for static code checks (in SonarCube) for it.

eaglerainbow:
Given that we have https://github.com/promregator/promregator/wiki/Simplification---Start-for-V1.0 in the meantime, I am currently thinking about whether it would be a better idea to merge this PR into `simplifiedProm` instead of `master`. We could be much more radical there.
What do you think?

src/test/java/org/cloudfoundry/promregator/endpoint/DiscoveryEndpointControllerTest.kt
```diff
@@ -0,0 +1,65 @@
...
+
+@ExtendWith(SpringExtension::class)
+@WebMvcTest(DiscoveryEndpoint::class, properties = ["promregator.simulation.enabled=true", "promregator.lifecycle.enabled=false"])
+class DiscoveryEndpointControllerTest(
+       @Autowired private val mockMvc: MockMvc
+) {
+    @MockkBean private lateinit var discoverer: CFMultiDiscoverer
+    @MockkBean(relaxed = true) private lateinit var config: PromregatorConfiguration
+
+    @BeforeEach
```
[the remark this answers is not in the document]
checketts: Nope. It is not required and can be deleted.

checketts:
I would be fine with just merging this into the `simplifiedProm` branch, especially since the wikipage indicates that `master` is end-of-life.

That being said, I've been digging into the tests and simplfying/improving them and I keep making those changes based on this branch. So that will focus my efforts into the `simplifiedProm` branch as well. Which is good all around, in my opinion.

eaglerainbow:
>  especially since the wikipage indicates that master is end-of-life.

I would not put it to that extreme. The idea for the branch split here is that we two can do things more radical. If we succeed, the idea is to build a new v1.0.0 out of `simplifiedProm`. If we should fail (what I neither hope nor believe), we still have a clean `master`. And finally, if something ugly happens for the today-productive consumers, we have a clean and consistent codeline from where we know that we can work on.

checketts:
I rebased this against `simplifiedProm` and changed the merge target.

src/test/java/org/cloudfoundry/promregator/endpoint/SingleTargetMetricsEndpointTest.kt
```diff
@@ -0,0 +1,160 @@
...
+        assertThat(response).isNotEmpty()
+
+        val parser = Parser(response)
+        val mapMFS = parser.parse()
+
+        assertThat(mapMFS).containsKey("metric_unittestapp")
+        assertThat(mapMFS).doesNotContainKey("metric_unittestapp2")
+    }
+
+    @Disabled("This test is no longer needed since label enrichment is removed")
```
checketts: These tests are failing because they were looking for metrics that are added due to label enrichment. With that feature removed, it seems like we can remove these tests. Let me know if that is OK and I'll rip them out.

src/test/java/org/cloudfoundry/promregator/endpoint/EnabledEnrichmentSingleTargetMetricsEndpointTest.java
```diff
@@ -1,135 +0,0 @@
...
-import org.springframework.test.context.TestPropertySource;
-import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
-
-import io.prometheus.client.Collector.MetricFamilySamples;
-import io.prometheus.client.Collector.MetricFamilySamples.Sample;
-
-@RunWith(SpringJUnit4ClassRunner.class)
-@SpringBootTest(classes = LabelEnrichmentMockedMetricsEndpointSpringApplication.class)
-@TestPropertySource(locations="enabledLabelEnrichment.properties")
-public class EnabledEnrichmentSingleTargetMetricsEndpointTest {
```
checketts: Since enrichment is removed, this test will fail, and is now removed.

eaglerainbow: make sense.

src/test/java/org/cloudfoundry/promregator/endpoint/SingleTargetMetricsEndpointTest.kt
eaglerainbow: yep. I do agree. Let's toss them.

src/test/java/org/cloudfoundry/promregator/endpoint/SingleTargetMetricsEndpointTest.java
```diff
@@ -1,152 +0,0 @@
...
-import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
-
-import io.prometheus.client.Collector.MetricFamilySamples;
-import io.prometheus.client.Collector.MetricFamilySamples.Sample;
-
-@RunWith(SpringJUnit4ClassRunner.class)
-@SpringBootTest(classes = MockedMetricsEndpointSpringApplication.class)
-@TestPropertySource(locations="default.properties")
-@ActiveProfiles("SingleTargetMetricsEndpointTest")
-public class SingleTargetMetricsEndpointTest {
```
eaglerainbow: Note to myself: Yes, also this test may go, as we still have `SingleTargetMetricsEndpointTest.testGetMetrics`

checketts:
I removed the obsolete tests. What can I do to setup the Kotlin Sonar support?

eaglerainbow:
> I removed the obsolete tests. What can I do to setup the Kotlin Sonar support?

Mainly unclear to me as well. I just had a look at https://sonarcloud.io/documentation/branches/overview/ and I am also unsure whether this is "short-term" or "long-term" in their sense. My gut-feeling currently tends more towards "short-term". 

The current sonarcloud integration of the pipeline can be found [here](https://github.com/promregator/promregator/blob/a0a6f2b01c02dbf386fe265bf32cfa633940a3fc/Jenkinsfile#L59). Mind the usage of the credentials there as well. 
However, I know that the maven plugin called there makes use of the pre-computed data such as jacoco and pmd. I don't know how this is with kotlin (or which static source code scanning tools / coverage analyzers need to be installed there). 
As I am the only guy  having access to the build server right now, I fear that I have to dig in there (as I will be the only one being able to iterate). 

The alternative to this is to - for the sake of this PR - *not* to go for sonarQube. However, we should ensure that we don't forget about this.

What's your take on this?

eaglerainbow:
BTW: Just made a project card for this in our project....
see https://github.com/promregator/promregator/projects/2#card-34092678

pom.xml
eaglerainbow: decided to be postponed. Put a reminder at https://github.com/promregator/promregator/projects/2#card-34092678

PR #157, opened 2020-02-20
Simplify resolveTargets reactive code to make the flow more clear.

checketts:
## Current Situation
The current `resolveTargets` method stores intermediate variables for the sstagess of the flux. This makes the flow hard to follow.

## Problem Statement
By inlining the flow it becomes simpler too read and reason about. Also enabling parallel processing of the flux objects can yield improved performance.

## Solution Approach
To better understand the reactive flow of resolving targets, I reformatted the code. While doing so I added the `parallel` operator. This breaks the work up to be run in multiple threads. You can visualize the separate threads when you enable the corresponding debug logging.

This reduces the discovery (for me from 5-8 seconds down to 4-5 seconds) and lays the groundwork for also being able to use this benefit when discovering against multiple CF APIs (a forthcoming PR)

## Post Activities
After merging this in I plan on adding (in upcoming PRs):

* Remove the `block` call and make this method return a `Mono<List<ResolvedTargets>>` so discovery itself can use the reactive workflow
* Add in config support for calling against multiple CF APIs
* Add in support for Consul based discovery

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
+++ b/src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
@@ -20,6 +20,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
 
 public class ReactiveTargetResolver implements TargetResolver {
 	private static final Logger log = LoggerFactory.getLogger(ReactiveTargetResolver.class);
@@ -49,8 +50,37 @@ public class ReactiveTargetResolver implements TargetResolver {
 			this.resolvedApplicationName = source.resolvedApplicationName;
 			this.resolvedApplicationId = source.resolvedApplicationId;
 		}
-		
+
+		public IntermediateTarget(Target target) {
+			this.configTarget = target;
+		}
+
 		public ResolvedTarget toResolvedTarget() {
+			if (getResolvedOrgId() == null) {
+				log.error(String.format("Target '%s' created a ResolvedTarget without resolved org id", getConfigTarget()));
+			}
+
+			if (getResolvedSpaceId() == null) {
+				log.error(String.format("Target '%s' created a ResolvedTarget without resolved space id", getConfigTarget()));
+			}
+
+			if (getResolvedApplicationId() == null) {
+				log.error(String.format("Target '%s' created a ResolvedTarget without resolved application id", getConfigTarget()));
+			}
+
+			if (getResolvedOrgName() == null) {
+				log.error(String.format("Target '%s' created a ResolvedTarget without resolved org name", getConfigTarget()));
+			}
+
+			if (getResolvedSpaceName() == null) {
+				log.error(String.format("Target '%s' created a ResolvedTarget without resolved space name", getConfigTarget()));
+			}
+
+			if (getResolvedApplicationName() == null) {
+				log.error(String.format("Target '%s' created a ResolvedTarget without resolved application name", getConfigTarget()));
+			}
+
+
 			ResolvedTarget rt = new ResolvedTarget();
 
 			rt.setOriginalTarget(this.configTarget);
@@ -174,57 +204,26 @@ public class ReactiveTargetResolver implements TargetResolver {
 		
 		
 	}
-	
+
 	@Override
 	public List<ResolvedTarget> resolveTargets(List<Target> configTargets) {
-		
-		Flux<IntermediateTarget> initialFlux = Flux.fromIterable(configTargets)
-				.map(configTarget -> {
-					IntermediateTarget it = new IntermediateTarget();
-					it.setConfigTarget(configTarget);
-					
-					return it;
-				});
-		
-		Flux<IntermediateTarget> orgResolvedFlux = initialFlux.flatMap(this::resolveOrg).log(log.getName()+".resolveOrg");
-		Flux<IntermediateTarget> spaceResolvedFlux = orgResolvedFlux.flatMap(this::resolveSpace).log(log.getName()+".resolveSpace");
-		Flux<IntermediateTarget> applicationResolvedFlux = spaceResolvedFlux.flatMap(this::resolveApplication).log(log.getName()+".resolveApplication")
-				.doOnNext(it -> {
-					if (it.getResolvedOrgId() == null) {
-						log.error(String.format("Target '%s' created a ResolvedTarget without resolved org id", it.getConfigTarget()));
-					}
-					
-					if (it.getResolvedSpaceId() == null) {
-						log.error(String.format("Target '%s' created a ResolvedTarget without resolved space id", it.getConfigTarget()));
-					}
-					
-					if (it.getResolvedApplicationId() == null) {
-						log.error(String.format("Target '%s' created a ResolvedTarget without resolved application id", it.getConfigTarget()));
-					}
-					
-					if (it.getResolvedOrgName() == null) {
-						log.error(String.format("Target '%s' created a ResolvedTarget without resolved org name", it.getConfigTarget()));
-					}
-					
-					if (it.getResolvedSpaceName() == null) {
-						log.error(String.format("Target '%s' created a ResolvedTarget without resolved space name", it.getConfigTarget()));
-					}
-					
-					if (it.getResolvedApplicationName() == null) {
-						log.error(String.format("Target '%s' created a ResolvedTarget without resolved application name", it.getConfigTarget()));
-					}
-				});
-		
-		Flux<ResolvedTarget> resultFlux = applicationResolvedFlux.map(IntermediateTarget::toResolvedTarget);
-		
-		List<ResolvedTarget> resultList = resultFlux.distinct().collectList().block();
-		if (log.isDebugEnabled()) {
-			log.debug(String.format("Successfully resolved %d configuration targets to %d resolved targets", configTargets.size(), resultList.size()));
-		}
-		
-		return resultList;
+		return Flux.fromIterable(configTargets)
+				.parallel()
+				.runOn(Schedulers.parallel())
+				.map(IntermediateTarget::new)
+				.flatMap(this::resolveOrg)
+				.log(log.getName() + ".resolveOrg")
+				.flatMap(this::resolveSpace)
+				.log(log.getName() + ".resolveSpace")
+				.flatMap (this::resolveApplication)
+				.log(log.getName() + ".resolveApplication")
+				.map(IntermediateTarget::toResolvedTarget)
+				.sequential()
+				.distinct().collectList()
+				.doOnNext(it -> log.debug("Successfully resolved {} configuration targets to {} resolved targets", configTargets.size(), it.size()))
+				.block();
 	}
-	
+
 	private Flux<IntermediateTarget> resolveOrg(IntermediateTarget it) {
 		/* NB: Now we have to consider three cases:
 		 * Case 1: both orgName and orgRegex is empty => select all orgs


src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
```diff
@@ -49,8 +50,37 @@ public IntermediateTarget(IntermediateTarget source) {
 			this.resolvedApplicationName = source.resolvedApplicationName;
 			this.resolvedApplicationId = source.resolvedApplicationId;
 		}
-		
+
+		public IntermediateTarget(Target target) {
+			this.configTarget = target;
```
checketts: Added new constructor to remove this work from the `resolveTarget` flow and reduce noisses

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
```diff
@@ -49,8 +50,37 @@ public IntermediateTarget(IntermediateTarget source) {
...
+				log.error(String.format("Target '%s' created a ResolvedTarget without resolved org name", getConfigTarget()));
+			}
+
+			if (getResolvedSpaceName() == null) {
+				log.error(String.format("Target '%s' created a ResolvedTarget without resolved space name", getConfigTarget()));
+			}
+
+			if (getResolvedApplicationName() == null) {
+				log.error(String.format("Target '%s' created a ResolvedTarget without resolved application name", getConfigTarget()));
+			}
```
checketts: Moved this logging out of the `resolveTarget` flow to reduce noise there.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
```diff
@@ -174,57 +204,26 @@ public void setResolvedApplicationId(String resolvedApplicationId) {
...
-		
-		Flux<ResolvedTarget> resultFlux = applicationResolvedFlux.map(IntermediateTarget::toResolvedTarget);
-		
-		List<ResolvedTarget> resultList = resultFlux.distinct().collectList().block();
-		if (log.isDebugEnabled()) {
-			log.debug(String.format("Successfully resolved %d configuration targets to %d resolved targets", configTargets.size(), resultList.size()));
-		}
-		
-		return resultList;
+		return Flux.fromIterable(configTargets)
```
checketts: This logic is identical to the previous logic, but uses the fluent chaining to make it simpler to follow. 

Also adds in the `parallel` support. This gives a 30% improvement in my current simple use cases. And based on when this supports calling multiple CF APIs as much as a 75% improvement (based on Promregator-lite behavior).

How it works:
`parallel` splits up the work (the flux) and `runOn(Schedulers.parallel())` actually assigns it to the separate threads with a count equal to the number of CPUs available by default.  Then `sequential` joins the paarallel work back together into a normal flux again.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
```diff
@@ -174,57 +204,26 @@ public void setResolvedApplicationId(String resolvedApplicationId) {
...
+				.log(log.getName() + ".resolveOrg")
+				.flatMap(intTarget -> resolveSpace(intTarget).parallel().runOn(Schedulers.parallel()))
+				.log(log.getName() + ".resolveSpace")
+				.flatMap (intTarget -> resolveApplication(intTarget).parallel().runOn(Schedulers.parallel()))
+				.log(log.getName() + ".resolveApplication")
+				.map(IntermediateTarget::toResolvedTarget)
+				.sequential()
+				.distinct().collectList()
+				.doOnNext(it -> log.debug("Successfully resolved {} configuration targets to {} resolved targets", configTargets.size(), it.size()))
+				.block();
```
checketts: Looking forward to removing this `block`, but will save that work for an upcoming PR.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
eaglerainbow: I understood that point.
However, it is unclear to me why you need to issue `parallel()` execution several times here. As far as I had observed it, once in a ParallelFlux, all your subsequent commands are distributed to the rails.

checketts: The subsequent `parallel` calls are to make the org/space/application lookups also parallel.

For example. Let's say I have 3 config targets, and 4 CPUs. The 3 config targets would each be assigned a single rail, but as the first config target begins resolving org and space info, all those requests would be running one at a time in that first rail. Does that make sense?

eaglerainbow: Not yet, but that may also be due to the fact that it's already late here (local time): As `resolveSpace` depends on the result of `resolveOrg` (and `resolveApplication` vs. `resolveSpace` likewise), in your "only-three-config-targets" case the maximal parallelization factor we can achieve is 3. 

I would only understand your point, if the first rail would grab the complete set of items in the flux when calling `next()` initially. But then we should try doing something against that instead of "hoping" that the next `parallel()` will eat smaller chunks.

Need to both think about this case a little more + play around with it...

checketts: This is a bit hard to describe. I tried creating a chart and a marble diagram, but that didn't help.

We aren't 'hoping' that each parallel grabs smaller chunks, however the explanation is a little harder due to the fact that `resolveOrg`/`resolveSpace`/`resolveApplication` each have 3 cases they are supporting. So I'm going to describe this as though the config target is using a regex for the org, leaving the space empty, and using a regex for application names, since that is the most expensive because it requires listing all orgs/spaces/apps to check the regex. Also it is the most power and simplest to configure for.

# No parallel

With no `parallel` then we march through the configTargets one at a time, and through each `flatMap` one at a time. The multiple of `configTargets` won't matter that much since the calls will be cached for `configTarget`s 2 and 3.

# Single Parallel call (on outer most Flux, line 211 in this code)

This case also doesn't add a lot of improvement, since we get the most benefit when the flux where we have the greatest number of elements. In CloudFoundry, that is probably in listing Orgs or Spaces.

When target1 begins it resolves orgs and waits for a response, target2 receives the same flux (since it was `cached()`(in prom-lite) or waits on the `Mono.fromFuture` (in Promregator's Caffeine cache). Either way, the org fluxes get resolved by the target1 rail, and target2, just follows in lockstep. The same happens for `resolveSpace` and `resolveApplication`. There isn't a chance for the target2 rail to do work that isn't already being done on the target1 rail.

# Parallel on outer and also within each `flatMap` call

Multiple `parallel` calls aren't bad because we aren't starting a new threadpool for each `parallel` call. Instead it just breaks up the work. So there isn't an increased cost but we get the benefit of doing multiple pieces of work at the same time.

When target1 begins resolving orgs it waits on the `retrieveAllOrgIds` call. target2 will also wait on that call. Again, in lockstep. Once it is resolved, the regex will likely match just a single org. So still no large benefit. If the regex matched multiple orgs, we would start seeing a benefit, because each org would be getting split off onto a rail to begin work. The same benefit happens below with spaces, so I'll revisit it there.

Now when target1 begins resolving the spaces, it calls `retrieveSpaceIdsInOrg` and if we assume only a single org had matched before, then target2 would reach the same call and be waiting for the result to return (since they both need all spaces). When the space lookup resolves, let's assume it is a flux with 40 elements. The `flatMap` is being returned, after making a single call, so in theory, there isn't any more work to be done. However by calling `parallel` in the `resolveSpace` `flatMap` it assigns out the returned fluxes to rails.

Now when target1 calls resolveApplication for that space, it inherits the rail that the space flux was assigned to. This means that when `retrieveAllApplicationIdsInSpace` is awaiting a result for spaceA, it can be making the request for spaceB, spaceC, and spaceD at the same time.

So in the example I described, you get the benefit as the calls below the `parallel` call fan out. So in my usecase I described, you could get the majority of the benefit by putting the `parallel` call on the `resolveSpace` line. However since there is no memory cost to assigning each group of fluxes to rails, we can have the most flexbility and concurrency by splitting out all fluxes to make the `parallel`.

I would love another rxjava user to double check my reasoning (sorry for the long winded response!).

To visualize the parallelism, you can enable debug logging, and view the thread name as the calls are being processed:
```
logging:
  level:
    org:
      cloudfoundry:
        promregator: DEBUG
```

eaglerainbow: I think I understood what you mean, but I am a little confused that this should be the case.
Let me check some things here first...

eaglerainbow: I think I figured out the point: Caffeine uses Futures internally, which are not backpressure-enabled. This effectively blocks the thread. The classical cache was much better in this regard, as it also worked with `Mono`s internally. In that sense the blocking behavior of the two cache types differ ==> 💩 

On the other hand: Good catch, @checketts !
(Still thinking what is the right approach here...)

checketts: We could change the Caffeine Futures to just cache the Monos directly (that is how I've been running with prom-lite). I know you ran into trouble with that in the past, but I wonder if it was related to the recently fixed CF client fix.

eaglerainbow: I just did a small experiment, but that did not confirm my claim above:

I added the unit test
```java
	@Test
	public void testBlockingBehavior() throws InterruptedException {
		FakeTicker ticker = new FakeTicker();
		
		AsyncLoadingCache<String, Integer> subject = Caffeine.newBuilder()
				.expireAfterAccess(240, TimeUnit.SECONDS)
				.refreshAfterWrite(120, TimeUnit.SECONDS)
				.ticker(ticker::read)
				.recordStats()
				.buildAsync(new AsyncCacheLoaderTimingImplementation());
		
		
		Flux<Integer> f = Flux.range(1, 10)
				.flatMap(i -> {
					@NonNull CompletableFuture<Integer> cf = subject.get(i+"");
					return Mono.fromFuture(cf);
				});
		f.doOnEach(i -> {
			log.info(String.format("Received %d", i.get()));
		}).blockLast();
	}
``` 

to https://github.com/promregator/promregator/blob/b55cc3479837fcb1fe5e096d0cb097cd1fb25c4a/src/test/java/org/cloudfoundry/promregator/cache/CaffeineAsyncLoadingTest.java#L22 . This is very much what happens here in this PR (without the `parallel` stuff). The logging printout (including timestamps and tasks) is:
```
21:40:31.337 [main] DEBUG reactor.util.Loggers$LoggerFactory - Using Slf4j logging framework
21:40:31.357 [main] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Request loading iteration 0 for request 1
21:40:31.406 [main] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Received 0
21:40:31.407 [main] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Request loading iteration 1 for request 2
21:40:31.412 [ForkJoinPool.commonPool-worker-1] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Starting to delay - iteration: 1
21:40:31.412 [main] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Request loading iteration 2 for request 3
21:40:31.412 [ForkJoinPool.commonPool-worker-2] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Starting to delay - iteration: 2
21:40:31.412 [main] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Request loading iteration 3 for request 4
21:40:31.412 [ForkJoinPool.commonPool-worker-1] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Starting to delay - iteration: 3
21:40:31.412 [main] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Request loading iteration 4 for request 5
21:40:31.413 [ForkJoinPool.commonPool-worker-3] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Starting to delay - iteration: 4
21:40:31.413 [main] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Request loading iteration 5 for request 6
21:40:31.413 [ForkJoinPool.commonPool-worker-1] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Starting to delay - iteration: 5
21:40:31.413 [main] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Request loading iteration 6 for request 7
21:40:31.413 [ForkJoinPool.commonPool-worker-1] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Starting to delay - iteration: 6
21:40:31.413 [main] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Request loading iteration 7 for request 8
21:40:31.413 [ForkJoinPool.commonPool-worker-1] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Starting to delay - iteration: 7
21:40:31.413 [main] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Request loading iteration 8 for request 9
21:40:31.413 [ForkJoinPool.commonPool-worker-1] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Starting to delay - iteration: 8
21:40:31.414 [main] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Request loading iteration 9 for request 10
21:40:31.414 [ForkJoinPool.commonPool-worker-1] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Starting to delay - iteration: 9
21:40:31.614 [parallel-2] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Finished delaying - iteration: 3
21:40:31.614 [parallel-4] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Finished delaying - iteration: 5
21:40:31.614 [parallel-3] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Finished delaying - iteration: 4
21:40:31.614 [parallel-5] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Finished delaying - iteration: 6
21:40:31.614 [parallel-1] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Finished delaying - iteration: 2
21:40:31.614 [parallel-2] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Received 3
21:40:31.614 [parallel-2] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Received 5
21:40:31.614 [parallel-1] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Finished delaying - iteration: 1
21:40:31.614 [parallel-2] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Received 6
21:40:31.614 [parallel-2] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Received 4
21:40:31.614 [parallel-2] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Received 1
21:40:31.614 [parallel-2] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Received 2
21:40:31.615 [parallel-6] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Finished delaying - iteration: 7
21:40:31.615 [parallel-8] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Finished delaying - iteration: 9
21:40:31.615 [parallel-7] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Finished delaying - iteration: 8
21:40:31.615 [parallel-6] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Received 7
21:40:31.615 [parallel-6] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Received 8
21:40:31.615 [parallel-6] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Received 9
21:40:31.616 [parallel-6] INFO org.cloudfoundry.promregator.cache.CaffeineAsyncLoadingTest - Received null
```

From the timestamps around `21:40:31.413` on the `[ForkJoinPool.commonPool-worker-1]` thread, we see that the `delayElement()` frees the thread. The same also happens at cf-java-client, if we are talking to the CFCC. The thread `[main]` is also returned back control as we can see that within the same millisecond further items are processed (`main` here stands for a single rail). 

Furthermore, we can also see that the results (upon reception) are automatically parallelized (thread name `parallel-*` -- I am here at a machine with 8 cores...).
I take out of this that even if we don't have backpressure, `Mono.fromFuture` is good enough to manage threading properly. 
==> my assumption is above is wrong (i.e. there must be another reason for the behavior you are observing).

checketts: Hmm. You are right. Thanks for sharing that unit test. I tried altering it in many ways and the only changes that stood out to me was the we do a `collectList().block()` instead of `blockLast()`, but even with that change it behaves as you described.

We also make additional requests based on previous requests. So I'll do some more digging to try to understand what is going on. I think your example that shows the `main` thread as a single rail helped me understand. Because I was seeing the `resolve.org` and similar log messages on given thread, I assumed that is where the work was happening, but instead that is merely the thread initiating the work.

Would you like me to remove the `parallel()` calls from this PR and create a separate PR for that purpose?

eaglerainbow: Thanks for considering all this! However, [you claim](https://github.com/promregator/promregator/pull/157#discussion_r381741816) that you saw up to 30% of improvement with this approach. The fact that I found a unit test which lets [the sequence of arguments](https://github.com/promregator/promregator/pull/157#discussion_r382317858) look strange does not mean that you cannot gain efficiency with it. It only means that another explanation might be necessary. 

Could you cross-check on how you perform the efficiency measurements? Perhaps this coding had influence on something else as well? How did you measure the gain (in e2e latency)? Or perhaps given this insight these measurements also become questionable?
Only if the latter is the case ....

> Would you like me to remove the parallel() calls from this PR and create a separate PR for that purpose?

... I'd like to kindly ask you to remove those `parallel()` statements. If we really can gain something, we should understand why and how. Because then it might also mean that we have to apply the same pattern elsewhere. In case we come to the conclusion that it won't help much, we should refrain from increasing the code's complexity.

Does this make sense to you?

checketts: At this point. I'm going to remove the `parallel` calls. It doesn't change the structure of the code, so it will be simple to add back in as I do further testing.

My method of measuring performance was making a request to the `/discovery` endpoint in my browser and noting it was taking 20 seconds before parallelizing, then it began to take 7 seconds after making it parallel. However, it was only in one 20 minute window. So I'll revisit it to make sure I can reproduce the performance benefit.

eaglerainbow: That makes sense. I could understand this benefit could already be reached by the `parallel()` in line 211 (to which I am not objecting at all, btw).

checketts: Pushed the simplified code. I left only the outer most `parallel` call.

eaglerainbow: Yep, just saw it. I'll trigger the voter. Let's see if it is also okay with this.

PR #160, opened 2020-02-26
Added logic to remove duplicated labels

### LINKED ISSUE #159: Duplicate labeling with label enrichment causes prometheus to not scrape
## Summary / Problem Statement
In prometheus = 3.16 when a metrics has duplicated labels prometheus marks the target as down and doesn't scrape it.  When enriched labels are used there is no way to override these values from the service being scraped (eg: app_name).  Enriched labels can be turned off, but not all applications may want to override these labels.

## Observed Behavior
Prometheus marks promregator as down and won't scrape due to one metric having a duplicate label in prometheus = 3.16

## Expected Behavior
Promregator adds a given label from enriched labels only if the label doesn't currently exist.  I was able to modify this locally to behave this way, and would be happy to put in a pr.

mcsammac:
## Current Situation
Currently if a metric has a label that is also provided by the label enrichment.  The metric ends up with a duplicated label.

## Problem Statement
In newer version of prometheus this causes the prometheus process to consume a high amount of CPU, and is blocked all together by the newest version of prometheus.  See issue: https://github.com/prometheus/prometheus/issues/6655

## Solution Approach
If a metric already has a label that would be added via label enrichment, let the metrics original label take priority and not add that label via label enrichment.

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/rewrite/AbstractMetricFamilySamplesEnricher.java
+++ b/src/main/java/org/cloudfoundry/promregator/rewrite/AbstractMetricFamilySamplesEnricher.java
@@ -48,9 +48,7 @@ public abstract class AbstractMetricFamilySamplesEnricher {
 	}
 	
 	protected abstract List<String> getEnrichedLabelNames(List<String> original);
-	
-	public abstract List<String> getEnrichedLabelValues(List<String> original);
 
-	protected abstract List<String> getEnrichedLabelValues(List<String> originalLabelNames, List<String> originalLabelValues);
+	public abstract List<String> getEnrichedLabelValues(List<String> originalLabelNames, List<String> originalLabelValues);
 
 }
--- a/src/main/java/org/cloudfoundry/promregator/rewrite/AbstractMetricFamilySamplesEnricher.java
+++ b/src/main/java/org/cloudfoundry/promregator/rewrite/AbstractMetricFamilySamplesEnricher.java
@@ -27,11 +27,10 @@ public abstract class AbstractMetricFamilySamplesEnricher {
 			
 			List<Collector.MetricFamilySamples.Sample> newSamples = new LinkedList<>();
 			for (Collector.MetricFamilySamples.Sample sample : mfs.samples) {
-				this.removeDuplicateLabels(sample.labelNames);
 				Collector.MetricFamilySamples.Sample newSample = new Collector.MetricFamilySamples.Sample(
 						sample.name,
 						this.getEnrichedLabelNames(sample.labelNames),
-						this.getEnrichedLabelValues(sample.labelValues),
+						this.getEnrichedLabelValues(sample.labelNames, sample.labelValues),
 						sample.value);
 				newSamples.add(newSample);
 			}
@@ -52,6 +51,6 @@ public abstract class AbstractMetricFamilySamplesEnricher {
 	
 	public abstract List<String> getEnrichedLabelValues(List<String> original);
 
-	protected abstract void removeDuplicateLabels(List<String> original);
-	
+	protected abstract List<String> getEnrichedLabelValues(List<String> originalLabelNames, List<String> originalLabelValues);
+
 }
--- a/src/main/java/org/cloudfoundry/promregator/rewrite/AbstractMetricFamilySamplesEnricher.java
+++ b/src/main/java/org/cloudfoundry/promregator/rewrite/AbstractMetricFamilySamplesEnricher.java
@@ -47,7 +47,7 @@ public abstract class AbstractMetricFamilySamplesEnricher {
 		return newMap;
 	}
 	
-	protected abstract List<String> getEnrichedLabelNames(List<String> original);
+	protected abstract List<String> getEnrichedLabelNames(List<String> originalLabelNames);
 
 	public abstract List<String> getEnrichedLabelValues(List<String> originalLabelNames, List<String> originalLabelValues);
 
--- a/src/main/java/org/cloudfoundry/promregator/rewrite/AbstractMetricFamilySamplesEnricher.java
+++ b/src/main/java/org/cloudfoundry/promregator/rewrite/AbstractMetricFamilySamplesEnricher.java
@@ -27,6 +27,7 @@ public abstract class AbstractMetricFamilySamplesEnricher {
 			
 			List<Collector.MetricFamilySamples.Sample> newSamples = new LinkedList<>();
 			for (Collector.MetricFamilySamples.Sample sample : mfs.samples) {
+				this.removeDuplicateLabels(sample.labelNames);
 				Collector.MetricFamilySamples.Sample newSample = new Collector.MetricFamilySamples.Sample(
 						sample.name,
 						this.getEnrichedLabelNames(sample.labelNames),
@@ -51,6 +52,6 @@ public abstract class AbstractMetricFamilySamplesEnricher {
 	
 	public abstract List<String> getEnrichedLabelValues(List<String> original);
 
-
-
+	protected abstract void removeDuplicateLabels(List<String> original);
+	
 }
--- a/src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
+++ b/src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
@@ -49,29 +49,64 @@ public class CFAllLabelsMetricFamilySamplesEnricher extends AbstractMetricFamily
 	protected List<String> getEnrichedLabelNames(List<String> original) {
 		List<String> clone = new LinkedList<>(original);
 		for(String label: labelNames) {
-			clone.remove(label);
-			//TODO: log that label was removed
+			if(clone.indexOf(label) < 0) {
+				clone.add(label);
+			} else {
+				//TODO: log that label already existed
+			}
 		}
-		Collections.addAll(clone, labelNames);
 		return clone;
 	}
 
 	@Override
 	public List<String> getEnrichedLabelValues(List<String> originalLabelNames, List<String> originalLabelValues) {
-		List<String> labelListClone = new LinkedList<>(originalLabelValues);
-		Set<String> originalLabelNamesSet = new HashSet<>(originalLabelNames);
+		List<String> clone = new LinkedList<>(originalLabelValues);
+		boolean orgNameExists = false;
+		boolean spaceNameExists = false;
+		boolean appNameExists = false;
+		boolean instanceIdExists = false;
+		boolean instanceNumberExists = false;
+
 		for(String label: labelNames) {
-			if(originalLabelNamesSet.contains(label)){
-				labelListClone.remove(originalLabelNames.indexOf(label));
-				//TODO: Log that value was removed
+			int index = originalLabelNames.indexOf(label);
+			if(index > -1){
+				switch (label){
+					case LABELNAME_ORGNAME:
+						clone.set(index, orgName);
+						orgNameExists = true;
+						//TODO: log that value already existed
+						break;
+					case LABELNAME_SPACENAME:
+						clone.set(index, spaceName);
+						spaceNameExists = true;
+						//TODO: log that value already existed
+						break;
+					case LABELNAME_APPNAME:
+						clone.set(index, appName);
+						appNameExists = true;
+						//TODO: log that value already existed
+						break;
+					case LABELNAME_INSTANCEID:
+						clone.set(index, instanceId);
+						instanceIdExists = true;
+						//TODO: log that value already existed
+						break;
+					case LABELNAME_INSTANCE_NUMBER:
+						clone.set(index, getInstanceFromInstanceId(instanceId));
+						instanceNumberExists = true;
+						//TODO: log that value already existed
+						break;
+				}
 			}
 		}
-		labelListClone.add(orgName);
-		labelListClone.add(spaceName);
-		labelListClone.add(appName);
-		labelListClone.add(instanceId);
-		labelListClone.add(getInstanceFromInstanceId(instanceId));
-		return labelListClone;
+
+		if(!orgNameExists) {clone.add(orgName);}
+		if(!spaceNameExists) {clone.add(spaceName);}
+		if(!appNameExists){clone.add(appName);}
+		if(!instanceIdExists){clone.add(instanceId);}
+		if(!instanceNumberExists){clone.add(getInstanceFromInstanceId(instanceId));}
+
+		return clone;
 	}
 
 	private static String getInstanceFromInstanceId(String instanceId) {
--- a/src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
+++ b/src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
@@ -1,8 +1,10 @@
 package org.cloudfoundry.promregator.rewrite;
 
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 
 /**
  * A MetricFamilySamplesEnricher which enriches the labels of metrics by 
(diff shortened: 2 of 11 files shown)


src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
```diff
@@ -24,43 +26,39 @@
...
-		clone.add(this.instanceId);
-		clone.add(getInstanceFromInstanceId(this.instanceId));
-		
+		Collections.addAll(clone, labelsAndValuesToAdd.values().toArray(new String[0]));
 		return clone;
 	}
+
+	@Override
+	protected void removeDuplicateLabels(List<String> original){
+		for(String label: original) {
```
checketts: It would be good to log a warning if this behavior is occurring.

src/test/java/org/cloudfoundry/promregator/rewrite/MetricFamilySamplesEnricherTest.java
```diff
@@ -104,4 +104,48 @@ public void testNullEnricherSimple() {
...
+		Assert.assertEquals("The first label should have been the labelName", "labelName", labelNames[1]);
+		Assert.assertEquals("The first label should have been the testOrgName", "org_name", labelNames[2]);
+		Assert.assertEquals("The first label should have been the testSpaceName", "space_name", labelNames[3]);
+		Assert.assertEquals("The first label should have been the testComponent", "cf_instance_id", labelNames[4]);
+		Assert.assertEquals("The first label should have been the testComponent", "cf_instance_number", labelNames[5]);
+
+		List<String> labelValuesList = testSample.labelValues;
+		Assert.assertEquals("There should be exactly 6 values",6, labelValuesList.size());
+		String[] labelValues = labelValuesList.toArray(new String[0]);
+		Assert.assertEquals("The app_name should not have been empty or overridden", "original_appName", labelValues[0]);
```
checketts: I actually would prefer the app_name be overridden by Promregator, so an app isn't in the place too 'spoof' another app's identity/metrics.

src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
```diff
@@ -24,43 +26,39 @@
...
+			labelsAndValuesToAdd.remove(label);
```
eaglerainbow: Changing the state of the Enricher could be dangerous - for instance due to the coding here https://github.com/promregator/promregator/blob/b55cc3479837fcb1fe5e096d0cb097cd1fb25c4a/src/main/java/org/cloudfoundry/promregator/endpoint/SingleTargetMetricsEndpoint.java#L100

Most likely this won't cause any practical issues by chance, but the approach feels like "asking for trouble" to me.

Perhaps here is a better idea:
* in `getEnrichedLabelNames` check the `original` whether the label's name is already provided there. If yes, skip it (and don't add it to the result, if the name is already there)
* do the same kind of check in `getEnrichedLabelValues` as well. For that, the interface must be extended, such that also the labelNames are handed in as parameters. However, that is safe, as we know all consumers (and can adjust them). 

Then we are also safe with these kind of "dirty cases" I mentioned above.

What do you think?

checketts (commented):
As @eaglerainbow mentioned, this sort of enrichment might be reaching an end of life so I want to imagine how we would solve this via a long term solution.

If label enrichment wasn't a thing, and an app was still adding its own `app_name` metric, then it would be likely have the config of in Prometheus of:

```
       target_label: app_name
```

Which would have the same behavior of overwriting the apps label with the Promregator determined one.

eaglerainbow (commented):
see comment below

src/test/java/org/cloudfoundry/promregator/rewrite/MetricFamilySamplesEnricherTest.java
eaglerainbow: That's also a valid point....

@mcsammac How does it look like in your practice: Is the data provided by the apps valid (and hence just redundant) or are these guys "little pretenders" 😁 ?

src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
eaglerainbow: @checketts 
However, if you get for every metric + each target + each scraping interval (default 15s!) an own logging entry, this could also be an operational nightmare.

eaglerainbow: Perhaps make this logging switchable via configuration option, which is at state "log it" by default?

checketts: Good point about the noise/quantity. Definitely would want to be able to mute it if it isn't of use.

In other codebases where this can occur, they just log it the first time at `WARN`, then at `DEBUG` for subsequent.

eaglerainbow: I could also live with that approach. However, those "magic log level changes" can get on your nerves as well.

Let's ask our customer 😁 
@mcsammac: Which variant would you prefer as a user?

BTW: in this case this wouldn't be really easy to implement properly. We must expect that multiple threads could switch the "remember my decision" toggle here - and we would have to apply synchronization, if we really want to do it properly (however, if we would spool out two or three additional one's too much due to a race condition --- who would care?)

mcsammac: I think i like the former of using a toggle.  Maybe could be tied to an overall logging level for the entire service?

src/test/java/org/cloudfoundry/promregator/rewrite/MetricFamilySamplesEnricherTest.java
mcsammac: The names are valid in a since that they represent the app, but it is not the same as the app name in PCF.  I originally was going to go down the route of giving the enriched labels name precedent, but I ultimately thought it would be better to give devs the ability to override if they saw a need too.

src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
mcsammac: Yea that is a good point.  I will try and implement it this way and add to this pr.

src/test/java/org/cloudfoundry/promregator/rewrite/MetricFamilySamplesEnricherTest.java
checketts: Sounds good to me

src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
checketts: Sounds good to me

eaglerainbow: Sounds good to me, too. 

@mcsammac: Will you give it a try or shall one of us do it?

Internal remark: General idea: create a special sublogger for https://github.com/promregator/promregator/blob/817d6410352c8ca764426544f8abceb4fac96b70/src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java. log with level `warn`. How to disable it then needs to be mentioned in the "logging" documentation.

src/test/java/org/cloudfoundry/promregator/rewrite/MetricFamilySamplesEnricherTest.java
eaglerainbow: You see me a little astonished. 
For the `appName` I still could get your way of thinking -- if they want to call their app `donkey` in their reporting instead, then fine... However, how do you want to make sure then that they don't overwrite the `orgName` and `spaceName` as well? By that they could create a huge mess not just at their own app, but also at their neighbor's. Or do you not have any "multi-tenancy"-like environment where this would play a role?

src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
eaglerainbow: perfect. Please feel free to get back to us in case you get stuck somewhere...

src/test/java/org/cloudfoundry/promregator/rewrite/MetricFamilySamplesEnricherTest.java
eaglerainbow: The more I think of this case, the more I am asking myself why they are overwriting `appName` in first place. Do they need to retain compatibility? Don't they like what they see? Did they not know that this was auto-added for them? Do we collide by chance with some automatic extractor they use? Are they able (but unwilling only) to choose a their own, custom label name or can't they change?

Do you have contact with some of these folks such that you could ask?

src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
mcsammac: I can attempt to implement it, and go from there.

src/test/java/org/cloudfoundry/promregator/rewrite/MetricFamilySamplesEnricherTest.java
mcsammac: In the particular case of our devs it was not intentional, but i could see a developer making a case for wanting to modify it.  Particularly for simplification of prometheus queries. I can also see the case for protecting devs from themselves.  Could do a configuration toggle here as well that is toggled to be not overridden by default, but that may be a bit overboard.  Thoughts?

checketts: I would recommend against letting them override it, so not even offering a config option. If a user had an actual need to alter a specific label, Prometheus relabeling would suffice so we don't need to implement anything to support even that usecase.

eaglerainbow: I initially also chewed on the toggle idea, but I fear that this could become a configuration nightmare. What if there are "some devs" out there who need to protection from themselves - and some others who "know what they are doing"... ? :dizzy_face:

> Particularly for simplification of prometheus queries

Do you see a reason why they can't be simplified using another (custom) label name for the value?

> Thoughts?

I am currently on the trip to forbid it (i.e. overwriting the label values). Almost any time I can think of, we are running in some risk here. 
Too bad that we didn't prefix all labels by Promregator initially with an own namespace (e.g. `promregator_appName` or so). That would have had the implicit tag attached "Keep your fingers off"... 😭 


Side note/internal remark: This all happens only in case label enrichment is used. We don't have such problems with label rewriting (at Prometheus), as the label name there is configured by the user.  ==> won't be a topic after "Promregator-simplification" anymore.

eaglerainbow: One more thing to consider: If we forbid it, what would be the section that we write this change to in the Changelog (when we released the next version with it)?
1. Prometheus de facto did an incompatible change in a patch (as their impact mainly was bug-like) for which we need to adapt now.
2. We would suppressing data from the applications and hence were also incompatible. 
3. The protection preventing to overwrite could also be sold as an improvement/enhancement. 

Talking about [semver](https://semver.org/), item 2 would call out for a major bump. Item 1  hints more towards a patch version increment. And item 3 speaks for a new minor version. 😵

mcsammac: That was also something i considered (changing the label values), but was afraid of having to reconfigure a bunch of dashboards.  I'm good with protecting the developers and the ones that have a specific use case using the prometheus relabeling.  So i purpose this:  

1. Check for any collisons with the enriched labels in getEnrichedLabelNames not adding the colliding label/s to the label list.
2. Doing the same check in getEnrichedLabelValues and not adding the corresponding value/s. 

Should it still log a warning if there is a collison, if it is noted as a feature?  And if so do we only log in one of the checks?

checketts: > what would be the section that we write this change to in the Changelog

A minor bump would call attention to the change, but since it was bug like anyways, it shouldn't have been being used. So I would be fine with minor or patch.

checketts: > Should it still log a warning if there is a collision

This should be a rare situation. So logging each time as a warning would be fine. Make it a sub-logger so it can be muted separately from the main logger.

Out of curiousity, why do you use label enrichment instead of Prometheus relabeling?

mcsammac: We have been using Promregator for nearly a year, and honestly we were running a very old version until very recently, when we had a team that had capacity to own it.  When was relabeling introduced? I don't know that it was in place when we first started using Promregator, or we were being lazy 😄

src/main/java/org/cloudfoundry/promregator/rewrite/AbstractMetricFamilySamplesEnricher.java
```diff
@@ -52,6 +51,6 @@ public AbstractMetricFamilySamplesEnricher() {
 	
 	public abstract List<String> getEnrichedLabelValues(List<String> original);
 
-	protected abstract void removeDuplicateLabels(List<String> original);
-	
+	protected abstract List<String> getEnrichedLabelValues(List<String> originalLabelNames, List<String> originalLabelValues);
```
eaglerainbow: Do you have any special reason why you doubled the method? Looks rather complicated to me this way...

src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
```diff
@@ -26,40 +27,56 @@
...
 	}
 
 	@Override
-	protected void removeDuplicateLabels(List<String> original){
-		for(String label: original) {
-			labelsAndValuesToAdd.remove(label);
+	protected List<String> getEnrichedLabelValues(List<String> originalLabelNames, List<String> originalLabelValues) {
+		List<String> clone = new LinkedList<>(originalLabelValues);
+		for(String label: labelNames) {
+			if(originalLabelNames.contains(label)){
```
eaglerainbow: This logic here is behaving bad on performance. Note that this is important here, as we need to enrich possibly hundreds (if not thousands) of metrics within a short timeframe.
Additionally `originalLabelNames` can also be a large in size. 

Consider using a precalculated `HashSet<String>` here. That could improve performance significantly.

eaglerainbow (commented):
Just some ideas how to make it happen more easily.

src/main/java/org/cloudfoundry/promregator/rewrite/AbstractMetricFamilySamplesEnricher.java
mcsammac: I was worried about doing unnecessary work for anywhere else this method is called and not used, but now that i look again it appears to only be the tests that call this method.  I will remove the overload

src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
mcsammac: Yea that is a good call!

src/main/java/org/cloudfoundry/promregator/rewrite/AbstractMetricFamilySamplesEnricher.java
mcsammac: Actually i was wrong.  There are other classes (getOwnTelemetry, etc) that call it as well.  The iteration, and contains (using the HashSet) shouldn't have an impact of the performance.

src/test/java/org/cloudfoundry/promregator/rewrite/MetricFamilySamplesEnricherTest.java
eaglerainbow: It was introduced with 5ecf6e5def2482c56d43f47f7e63969e27121b33 and is included since version 0.5.0. That's roughly 18 months back...
You may find documentation on this at https://github.com/promregator/promregator/blob/a0a6f2b01c02dbf386fe265bf32cfa633940a3fc/docs/documentation.md#label-rewriting.

src/main/java/org/cloudfoundry/promregator/rewrite/AbstractMetricFamilySamplesEnricher.java
```diff
@@ -48,9 +48,7 @@ public AbstractMetricFamilySamplesEnricher() {
 	}
 	
 	protected abstract List<String> getEnrichedLabelNames(List<String> original);
```
eaglerainbow: Just a minor side note: I would
```suggestion
	protected abstract List<String> getEnrichedLabelNames(List<String> originalNames);
```
to prevent any misunderstand right from the beginning.

src/main/java/org/cloudfoundry/promregator/rewrite/CFAllLabelsMetricFamilySamplesEnricher.java
```diff
@@ -24,44 +29,51 @@
...
-		clone.add(this.instanceId);
-		clone.add(getInstanceFromInstanceId(this.instanceId));
-		
-		return clone;
+	public List<String> getEnrichedLabelValues(List<String> originalLabelNames, List<String> originalLabelValues) {
+		List<String> labelListClone = new LinkedList<>(originalLabelValues);
+		Set<String> originalLabelNamesSet = new HashSet<>(originalLabelNames);
+		for(String label: labelNames) {
+			if(originalLabelNamesSet.contains(label)){
+				labelListClone.remove(originalLabelNames.indexOf(label));
```
eaglerainbow: This is a double-lookup - and you will get an `O(n²)` algorithm again with it. 

BTW: Why are you removing it? It will be added anyhow afterwards. And the order of labels will be ignored by Prometheus. 

Perhaps an alternative idea, therefore, could be to "remember" (e.g. with five boolean variables) which labels you have already seen. Then, you could change the label's value directly in place - even without removing the value from the list before.

Just some thoughts...

src/main/java/org/cloudfoundry/promregator/rewrite/NullMetricFamilySamplesEnricher.java
```diff
@@ -23,10 +23,8 @@ public NullMetricFamilySamplesEnricher() {
 	protected List<String> getEnrichedLabelNames(List<String> original) {
 		return new LinkedList<>(original);
 	}
-	
+
 	@Override
-	public List<String> getEnrichedLabelValues(List<String> original) {
-		return new LinkedList<>(original);
-	}
+	public List<String> getEnrichedLabelValues(List<String> originalLabelNames, List<String> originalLabelValues) { return new LinkedList<>(originalLabelValues); }
```
eaglerainbow: please check indentation here.

PR #165, opened 2020-03-01
Suppress logging of warning message, if no target found - #164

### LINKED ISSUE #164: Disable Warning Message(s) in case of Targets not being resolvable
## Environment
Promregator 0.7.3 on CF using docker.
Original trigger: @warlockza 

## Summary / Problem Statement
Consciously unprecise target configuration may leave cases where an org/space is empty (no apps installed), but still are configured as target. 
There are also other cases where a resolution isn't possible

## Observed Behavior
We create too many warnings. They may pose an operational problem.

## Expected Behavior
The user is able to disable these type of warnings explicitly. However, it shall be possible to still receive other warnings from `reactivetargetresolver`. 

## Log of discussion

```
Question
If I specify an org and space that exist
But it does not contain any apps
And I have specified an app name
How do you expect promregator to behave
[...]

I am seeing lots of errors in the logs

[...]

is it "Application id could not be found for org '%s', space '%s' and application '%s'. Check your configuration; skipping it for now"?

Yes, that

that's line 264 (of reactivetargetresolver)
```

eaglerainbow:
## Current Situation
There can be a large number of warning messages of type 

```
Application id could not be found for org '%s', space '%s' and application '%s'. Check your configuration; skipping it for now
```

if the target configuration includes empty-target definitions.

## Problem Statement
The log is flooded with these messages unnecessarily.

## Solution Approach
An additional configuration option `promregator.resolver.logging.empty.target` is introduced which permits switching off the warning messages.

## Post Activities
- [ ] Adding release notes

closes #164

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
+++ b/src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
@@ -17,6 +17,7 @@ import org.cloudfoundry.promregator.config.Target;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
@@ -28,6 +29,9 @@ public class ReactiveTargetResolver implements TargetResolver {
 	@Autowired
 	private CFAccessor cfAccessor;
 
+	@Value("${promregator.resolver.logging.empty.target:true}")
+	private boolean emptyResolutionIsLogged;
+	
 	private static class IntermediateTarget {
 		private Target configTarget;
 		private String resolvedOrgName;
@@ -352,7 +356,7 @@ public class ReactiveTargetResolver implements TargetResolver {
 					.filter(appResource -> appNameToSearchFor.equals(appResource.getEntity().getName().toLowerCase(Locale.ENGLISH)))
 					.single()
 					.doOnError(e -> {
-						if (e instanceof NoSuchElementException) {
+						if (e instanceof NoSuchElementException && this.emptyResolutionIsLogged) {
 							log.warn(String.format("Application id could not be found for org '%s', space '%s' and application '%s'. Check your configuration; skipping it for now", it.getResolvedOrgName(), it.getResolvedSpaceName(), it.getConfigTarget().getApplicationName()));
 						}
 					})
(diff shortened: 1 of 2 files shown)


src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
```diff
@@ -352,7 +356,7 @@ public void setResolvedApplicationId(String resolvedApplicationId) {
 					.filter(appResource -> appNameToSearchFor.equals(appResource.getEntity().getName().toLowerCase(Locale.ENGLISH)))
 					.single()
 					.doOnError(e -> {
-						if (e instanceof NoSuchElementException) {
+						if (e instanceof NoSuchElementException && this.emptyResolutionIsLogged) {
 							log.warn(String.format("Application id could not be found for org '%s', space '%s' and application '%s'. Check your configuration; skipping it for now", it.getResolvedOrgName(), it.getResolvedSpaceName(), it.getConfigTarget().getApplicationName()));
```
checketts: y not just create a sub-logger that the user could disable via normal logging config?

eaglerainbow: To be honest, I was unsure about that approach myself. 
I had thought about this, but the configuration open then would look very strange:

```
logging.org.cloudfoundry.promregator.scanner.ReactiveTargetResolver.EmptyTargets=ERROR
```

It is simply very hard to explain in documentation.

The other would have been to keep the configuration approach as it is right now, but then to have a sublogger, whose level then needs to be tweaked during initialization. However that could also be wrong, because of someone then writes 

```
logging.org.cloudfoundry.promregator.scanner.ReactiveTargetResolver.EmptyTargets=DEBUG
```

into his config, then we also have conflicting statements there. 

So, this all wasn't really convincing to.
Or what do you think?

checketts: I would have it log at `info` level and if the user wishes to mute it they will most likely fix the config to not scan for that app by name (our ideal solution), but if they actually want to keep scanning for the missing target, they will noise in the logs and look for the typical way they would mute logging (controlling the logger level), by setting it to WARN or ERROR.

I suspect they wouldn't even think to look for the `promregator.resolver.logging.empty.target` property.

So my recommendation would be to not include `promregator.resolver.logging.empty.target`, but even if it is included, also separate out the noisy logger so it can be muted via typical means, without muting the entire class.

eaglerainbow: True that: People will be coming from the logger's name, not from the configuration page. You convinced me - I'll change it.

PR #167, opened 2020-03-04
Convert Promregator into a Kotlin data class and migrate all config properties into it

checketts:


eaglerainbow:
see also review comments at https://github.com/checketts/promregator/pull/1

src/test/java/org/cloudfoundry/promregator/endpoint/DiscoveryEndpointTest.kt
```diff
@@ -0,0 +1,93 @@
...
+            assertThat(applicationId).isEqualTo("faedbb0a-2273-4cb4-a659-bd31331f7daf")
+            assertThat(applicationName).isEqualTo("unittestapp")
+            assertThat(instanceId).isEqualTo("faedbb0a-2273-4cb4-a659-bd31331f7daf:0")
+            assertThat(instanceNumber).isEqualTo("0")
+            assertThat(orgName).isEqualTo("unittestorg")
+            assertThat(spaceName).isEqualTo("unittestspace")
+            assertThat(targetPath).isEqualTo("/singleTargetMetrics/faedbb0a-2273-4cb4-a659-bd31331f7daf/0")
+        }
+
+        with(response[1].labels) {
```
checketts: Using Kotlin [scope function](https://kotlinlang.org/docs/reference/scope-functions.html) to simplify the assertions.  `with` just changes the scope of `this` so we can do `assertThat(applicationId)` instead of `assertThat(response[1].labels.applicationId)`

src/test/java/org/cloudfoundry/promregator/endpoint/InvalidateCacheEndpointTest.kt
```diff
@@ -0,0 +1,47 @@
...
+import org.cloudfoundry.promregator.scanner.CachingTargetResolver
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.http.HttpStatus
+
+//@SpringBootTest(classes = MockedAppInstanceScannerEndpointSpringApplication.class)
+class InvalidateCacheEndpointTest {
+    private lateinit var subject: InvalidateCacheEndpoint
+    private val cfAccessorCache = mockk<CFAccessorCache>(relaxed = true)
+    private val targetResolver = mockk<CachingTargetResolver>(relaxed = true)
```
checketts: We can keep using Mockito for mocking in tests, but mockk brings a 'strict' mocking mode (along with some other great features). 

When a mock is called that is strict, if a stub hasn't been created, then we get an exception (with useful info about what was called, so you can easily mock it out). With Mockito, instead you get surprising null pointers.

In this test since we are just verifying, we are OK with the relaxed mode.

PR #172, opened 2020-03-09
Don't pass 'null' for all orgs.

checketts:
In #170 the user it trying to use an `orgRegex` and is seeing a null pointer.

This fixes by using a hard coded "all" instead of a hard coded `null`. This does not need to be backported in simplifiedProm since we use a different key there (the API)


eaglerainbow:
Good work - and I am in favor of merging.

The only question I have is: Why didn't we detect this before?

eaglerainbow:
BTW: Will still require double maint to `master`, though.
I'll take care of it after merging

eaglerainbow:
Just for the records: I have cross-checked that no other cache that would be fed with a hard-coded `null` key (if we had, we would have to apply the same pattern there as well).

eaglerainbow:
> The only question I have is: Why didn't we detect this before?

The simple but bad answer is: We are missing test coverage there...

eaglerainbow:
Just for the record: Double maintenance to `master` done with 1370accaffae3f27c9a5cc3732af2e8e5c1ede50.

checketts:
Thanks for adding the tests. I considered it and after seeing how simple it was for you to add, I should have done it :D

eaglerainbow:
No prob, dude. 😄 
One good turn deserves another.

PR #173, opened 2020-03-09
Add ability for partial result for discovery

checketts:
## Current Situation
As noted in #170, Promregator can pound the CFCC by sending too many requests. This is an idea for how we could support that.

## Problem Statement
Since targets with regexes can fan out into many requests. 

## Solution Approach
This solution keeps track of the number of requests are being made to the CFCC so we could configure a limit on them. Optionally, we might be able to use the total time the request is taking as well, but I didn't try that.  

A big change is we would move from the AsyncCacheLoader to a manual one. This allows us to determine if a request has already been cached and decide if we want to make it or not.

This was an investigative spike to discuss the approach more than a hardened solution

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeine.java
+++ b/src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeine.java
@@ -1,39 +1,34 @@
 package org.cloudfoundry.promregator.cfaccessor;
 
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.Executor;
-import java.util.concurrent.ScheduledThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
-
-import javax.annotation.PostConstruct;
-
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.Scheduler;
 import org.apache.log4j.Logger;
-import org.checkerframework.checker.nullness.qual.NonNull;
 import org.cloudfoundry.client.v2.applications.ListApplicationsResponse;
 import org.cloudfoundry.client.v2.organizations.ListOrganizationsResponse;
 import org.cloudfoundry.client.v2.spaces.GetSpaceSummaryResponse;
 import org.cloudfoundry.client.v2.spaces.ListSpacesResponse;
 import org.cloudfoundry.promregator.internalmetrics.InternalMetrics;
+import org.cloudfoundry.promregator.scanner.BackPressure;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
-
-import com.github.benmanes.caffeine.cache.AsyncCacheLoader;
-import com.github.benmanes.caffeine.cache.AsyncLoadingCache;
-import com.github.benmanes.caffeine.cache.Caffeine;
-import com.github.benmanes.caffeine.cache.Scheduler;
-
 import reactor.core.publisher.Mono;
-import reactor.core.scheduler.Schedulers;
+
+import javax.annotation.PostConstruct;
+import java.util.Optional;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
 
 public class CFAccessorCacheCaffeine implements CFAccessorCache {
 	private static final Logger log = Logger.getLogger(CFAccessorCacheCaffeine.class);
 
-	private AsyncLoadingCache<String, ListOrganizationsResponse> orgCache;
-	private AsyncLoadingCache<String, ListOrganizationsResponse> allOrgIdCache;
-	private AsyncLoadingCache<CacheKeySpace, ListSpacesResponse> spaceCache;
-	private AsyncLoadingCache<String, ListSpacesResponse> spaceIdInOrgCache;
-	private AsyncLoadingCache<CacheKeyAppsInSpace, ListApplicationsResponse> appsInSpaceCache;
-	private AsyncLoadingCache<String, GetSpaceSummaryResponse> spaceSummaryCache;
+	private Cache<String, Mono<ListOrganizationsResponse>> orgCache;
+	private Cache<String, Mono<ListOrganizationsResponse>> allOrgIdCache;
+	private Cache<CacheKeySpace, Mono<ListSpacesResponse>> spaceCache;
+	private Cache<String, Mono<ListSpacesResponse>> spaceIdInOrgCache;
+	private Cache<CacheKeyAppsInSpace, Mono<ListApplicationsResponse>> appsInSpaceCache;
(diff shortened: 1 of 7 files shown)


src/main/java/org/cloudfoundry/promregator/scanner/BackPressureOps.java
```diff
@@ -0,0 +1,53 @@
...
+                    }
+
+                    if (rootCause instanceof IncompleteResultException) {
+                        @SuppressWarnings("unchecked") Mono<List<T>> result = Mono.just(((IncompleteResultException) rootCause).getResults());
+                        return result;
+                    } else {
+                        return Mono.error(t);
+                    }
+                })
+                .subscriberContext(Context.of("backpressure", new BackPressure(maxRequests)));
```
checketts: This is the core feature. I was investigating using the Reactor `Context` api to be able to track the behavior.

src/main/java/org/cloudfoundry/promregator/scanner/BackPressureOps.java
```diff
@@ -0,0 +1,53 @@
...
+            int skips = bp.getRequestsSkipped();
+            LOG.info("Got results: " + results.size() + " backpressue: " + skips);
+            if (skips > 0) {
+                bp.reset();
+                throw new IncompleteResultException(skips, results);
+            }
+
+            return results;
+        }))
+                .retryBackoff(maxRetries, Duration.ofSeconds(1), Duration.ofSeconds(10))
```
checketts: If a discovery fails, we wait a bit (with some jitter) before retrying. This could be configurable. I don't think we would want to make it longer than 60 seconds though since that could start cutting into the next discovery.

eaglerainbow:
> This solution keeps track of the number of requests are being made to the CFCC so we could configure a limit on them.

Is there a special reason why you don't like Guava's RateLimiter? See also https://github.com/promregator/promregator/blob/34bcf77a55856fa946bd626994268840a2bd45d4/src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorRateLimit.java#L38 

On first glance, your approach looks quite complex to incorporate that into Reactor.

BTW: I have an updated variant, which brings the RateLimiter before the paging - and is even shorter in terms of code.

checketts:
The main reason I hadn't used the `RateLimiter` was was to investigate the 'total time approach' along with the capabilities of the Reactor Context capabilities. 

You are correct that my approach is much more complex than the `CFAccessorRateLimit`. I was hoping the Context API would be simpler to leverage, more like an MDC or ThreadLocal approach. I'm extremely new to it, so I wouldn't be surprised if I had mis-used it.

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeine.java
```diff
@@ -1,39 +1,34 @@
...
 public class CFAccessorCacheCaffeine implements CFAccessorCache {
 	private static final Logger log = Logger.getLogger(CFAccessorCacheCaffeine.class);
 
-	private AsyncLoadingCache<String, ListOrganizationsResponse> orgCache;
-	private AsyncLoadingCache<String, ListOrganizationsResponse> allOrgIdCache;
-	private AsyncLoadingCache<CacheKeySpace, ListSpacesResponse> spaceCache;
-	private AsyncLoadingCache<String, ListSpacesResponse> spaceIdInOrgCache;
-	private AsyncLoadingCache<CacheKeyAppsInSpace, ListApplicationsResponse> appsInSpaceCache;
-	private AsyncLoadingCache<String, GetSpaceSummaryResponse> spaceSummaryCache;
+	private Cache<String, Mono<ListOrganizationsResponse>> orgCache;
```
eaglerainbow: I don't really fancy making our caches synchronous. 
As far as I can get it tonight, this would mean that after expiry, the next request to the cache has to block until the value is fetched again. The advantageous Caffeine behavior would fall back more or less to the quality of the classic cache with this.

eaglerainbow:
I slowly get how you do it with the Context API. 
In essence what you want to do is to limit the number of requests we send.  To that idea, I am wondering, if it wouldn't be easier to compile a Flux of type "Generic request" containing all requests we want to send out (right now). Then we can do a `.limitRequest(long)` and/or `Flux#take(Duration)`.
However, such a cut-off could happen rather arbitrarily: 

Let's say we have 200 orgs with one space each. Each space has at least one app to scrape. The target config tells us `orgRegex: .*`. Then, the first request will be the org fetching. The requests no. 2 until no. 201 will be space requests. If we then have a limitRequest of - let's say - 150, then we will never be able to fetch any single application.

I am right that your `maxRequests` would potentially have the same issue?

checketts:
> I am right that your `maxRequests` would potentially have the same issue?

Yes. 

As I've thought about the `Context` approach, I'm less and less satisfied with it. Primarily since the context is based on a request, 2 Prometheus servers would double the load to the CFCC. I think centralized throttling using a ratelimiter is a better approach.

In diving into the Guava Ratelimiter one point I don't love is that the `acquire()` method is blocking. I still have a goal of being able to run a variant of Promregator via WebFlux and blocking the Netty threads should be avoided. Also it isn't possible to instrument for metrics (that I could tell), but that can be solved by instrumenting the RateLimiting accessor ourselves.

Since the `AsyncCacheLoader` happens in its own thread/executor that blocking isn't really a problem, just something I wanted to keep in mind.

I'm going to close this PR for now. I've learned lot about the Context strengths/limitations.

eaglerainbow:
> In diving into the Guava Ratelimiter one point I don't love is that the acquire() method is blocking

Me too. 
That is why I had to [pack them into a `fromCallable`](https://github.com/promregator/promregator/compare/metadataRateLimiter#diff-4278b14e403dbc7db74d903ff91dfa1dR50) and putting it onto [a `subscribeOn`](https://github.com/promregator/promregator/compare/metadataRateLimiter#diff-4278b14e403dbc7db74d903ff91dfa1dR73) which is the [recommended best practice for making blocking stuff integrated into reactor coding](https://projectreactor.io/docs/core/release/reference/#faq.wrap-blocking). If we had a non-blocking solution for rate limiting, I'd love to try using that instead. I am aware that the current approach isn't ideal as well: When throttling, the number of threads in blocked state may increase dramatically, as we are using an elastic scheduler. However, in @tlkamp's environment, the resources of both the number of threads and the memory consumption for the thread's stacks are not of concern. 

> I still have a goal of being able to run a variant of Promregator via WebFlux and blocking the Netty threads should be avoided.

In my solution they are not blocked: They will block an anonymous elastic thread. 

As the entire thing is packed into a Mono, I don't see the current approach to jeopardize your Webflux thinking. Where is your doubt?

> Also it isn't possible to instrument for metrics (that I could tell), but that can be solved by instrumenting the RateLimiting accessor ourselves. 

Yes, rather easily I would say. See my recent commits on this on `metadataRateLimiter`.

> (I might do a spike with the Resilience4J RateLimiter to compare its behavior)

Valid point. If there is something we can learn from there or if we can even change over to them and get instrumentation for free, I am open to that.

PR #186, opened 2020-10-01
Upgrade Spring-Boot version

tlkamp:
## Current Situation
 
Promregator is currently using an older version of Spring-Boot, which contains embedded Tomcat dependencies. These tomcat dependencies have vulnerabilities.

## Problem Statement
 
The embedded tomcat dependencies have vulnerabilities.

[CVE-2020-9484](https://nvd.nist.gov/vuln/detail/CVE-2020-9484)

## Solution Approach

I upgraded Spring-Boot to the next version that contains the embedded tomcat version that is not vulnerable. I verified this by running 

```console
$ mvn clean verify -DskipTests=true
$ mvn dependency:tree | grep tomcat
```

## Notes
- I could not determine if the existing version of Guava is sufficient. I checked [here](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-dependency-versions.html) and it mentions a version that is earlier than the one currently defined in the `pom.xml`. If this is the wrong place to check, let me know and I will also add some notes to the comments in the `pom.xml`.
- There are tests that are failing in the master branch that I started from, so those tests are failing here as well.

Happy Hacktoberfest!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ hs_err_pid*
 
 # files for intellij environment
 /.idea/
+*.iml
--- a/pom.xml
+++ b/pom.xml
@@ -246,7 +246,7 @@
 		<dependency>
 			<groupId>com.mikesamuel</groupId>
 			<artifactId>json-sanitizer</artifactId>
-			<version>1.2.0</version>
+			<version>1.2.1</version>
 		</dependency>
 
 		<dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,7 @@
 		
 		<spring-cloud.version>Hoxton.SR1</spring-cloud.version>
 		<prometheus.version>0.6.0</prometheus.version>
-		<cloudfoundry.client.version>3.24.0.RELEASE</cloudfoundry.client.version>
+		<cloudfoundry.client.version>3.27.0.RELEASE</cloudfoundry.client.version>
 		<logback.contrib.version>0.1.5</logback.contrib.version>
 
 		<sonar.projectKey>promregator</sonar.projectKey>
@@ -175,7 +175,7 @@
 		<dependency>
 			<groupId>com.google.guava</groupId>
 			<artifactId>guava-testlib</artifactId>
-			<version>28.1-jre</version>
+			<version>${guava-version}</version>
 			<scope>test</scope>
 		</dependency>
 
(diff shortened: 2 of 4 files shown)


tlkamp:
@eaglerainbow no problem, thank you!

eaglerainbow:
Just found time to have a look at this case:

* The Guava thing does not seem to be a problem: The main consumer which could collide would be the version of activemq-broker, but that is not affected by this bump. 
* The conditions required for CVE-2020-9484 are rather strict. I hardly can think of a situation where all of the conditions would hold true in a) a standard spring-boot application and b) in Promregator in particular. Therefore, yes, we should bump, but no, this isn't really a big threat.
* The unit test errors reported in the voter should also be the same unit tests which are failing on master. I have to have a second look at it what's wrong there... (will report back).

In sum: Looks good; we have some prerequisites to check though before we may merge.

eaglerainbow:
With bf0e2ac0b33065f474e257a872f8f03202f8defa I tweaked the unit tests a little. Perhaps this is not the super-solution to the problem, but I hope it will do the trick.
@tlkamp  Could you please merge/rebase this PR onto the new state of `master´? I ran a brief test afterwards together with your state, and it seems to work.

tlkamp:
> The conditions required for CVE-2020-9484 are rather strict. I hardly can think of a situation where all of the conditions would hold true in a) a standard spring-boot application and b) in Promregator in particular. Therefore, yes, we should bump, but no, this isn't really a big threat.

I agree. The vulnerability got flagged by my company, and we want to continue using Promregator. 😄 

I have merged/rebased your adjustments. Thank you!

eaglerainbow:
Ah, apparently you both rebased and merged against `master` :) This causes a little hiccup in the diff at Github.
Let's check if also the voter is able to digest that...

eaglerainbow:
Yep, that worked.... here we go...

And thanks again for your contribution!

tlkamp:
@eaglerainbow Thank you for working with me! Appreciate it. 

If I have other questions/ideas, what is the best way to get those to you? An issue?

eaglerainbow:
> An issue?

Yeah, most likely: We collect also discussions in issues tagging them with "discussion". 
If you wanna to talk to me more directly (e.g. because things are not ripe enough to go public), I can also invite you to the discord server I used with some of the folks (such as checketts).

Issue #190, opened 2020-12-24
need promregator.targets[].port to define specific port 8080 to fetch


ManuPin:
## Environment
I deploy promregator in a space of cloud foundry with versionning and manifest : 
---
applications:
  - name: promregator
    instances: 1
    memory: 1G
    disk_quota: 512M
    path: public/promregator-0.8.2.jar
    buildpacks:
      - java_buildpack_v4_35
    env:
      SPRING_APPLICATION_JSON: '{"promregator":{"authenticator":{"type":"none"},"targets":[{"orgName":"orgName","spaceName":"spaceName","path":"/actuator/prometheus","protocol":"http"}]},"cf":{"api_host":"api.cloudfoundry","username":"","password":""}}'

The api version:    2.147.0, And promeregator discover the app but can't fetch metric because can't find promregator.targets[].port parameter

## Expected Behavior
<!-- describe what behavior i woul expect a key value to define the port to fecth,in Subgroup "promregator.targets" , a promregator.targets[].port. Then we can define , protocol, path and port to fetch prometheus metrics for each target.

eaglerainbow:
Hi @ManuPin,

Thanks for joining in here! And thanks reporting your issue.

I am trying to understand your situation & context, but honestly I have a little trouble with. Here is what I got: 
* You are deploying Promregator to Cloudfoundry.
* You don't specify any routes for it (or just use the default route provided)
* Configuration happens using environment variable `SPRING_APPLICATION_JSON`. In that configuration you are defining a space to scrape. You want to make use of the HTTP protocol.
* Apparently, you intend to provide port 8080 within the configuration.

Note that there is no such thing as a `port` parameter to a target (see also https://github.com/promregator/promregator/blob/af3893bb2090fe10089ed1d14ddd43ebe62dabf0/docs/config.md#subgroup-promregatortargets ) -- and that is intended!
To explain that keep our [architecture picture](https://github.com/promregator/promregator/blob/af3893bb2090fe10089ed1d14ddd43ebe62dabf0/docs/architecture.png) in mind: Promregator will always be querying metrics going through the (go)routers of Cloud Foundry! Even if you deploy Promregator to CF itself, it does not make use of internal network communication. That is why I don't get the hang of it why you want to specify port 8080 in the configuration. Specifying it would mean that you try to connect to the haproxies of CF (in front of the gorouters) using port 8080... Promregator does not talk to your applications directly...

[...]

PR #194, opened 2021-01-22
WIP: Add internal domain support

ScottGuymer:
This is a Work in Progress to show my approach and get feedback.

## Current Situation

Promregator does not currently support routing for internal domains. This PR intends to add support for detecting an internal domain and providing the correct internal urls. See #190 for more info

## Problem Statement
I would like to use promregator with internal routes to keep metrics scraping private and use alternative ports on my apps for metrics servers.

## Solution Approach

- Add retrieval of the available domains from CF
- when detecting the URL for a target match it to a domain
- if the domain is internal use the internal url scheme
- use the port defined in the route for the scraping port

i looks like in the api when you retrieve routes for an app you can see the port that has been mapped there. Even though you have to explicitly use the port in the URL when using an internal route.

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessor.java
+++ b/src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessor.java
@@ -5,6 +5,7 @@ import org.cloudfoundry.client.v2.info.GetInfoResponse;
 import org.cloudfoundry.client.v2.organizations.ListOrganizationsResponse;
 import org.cloudfoundry.client.v2.spaces.GetSpaceSummaryResponse;
 import org.cloudfoundry.client.v2.spaces.ListSpacesResponse;
+import org.cloudfoundry.client.v3.domains.ListDomainsResponse;
 
 import reactor.core.publisher.Mono;
 
@@ -21,7 +22,9 @@ public interface CFAccessor {
 
 	Mono<ListApplicationsResponse> retrieveAllApplicationIdsInSpace(String orgId, String spaceId);
 
-	Mono<GetSpaceSummaryResponse> retrieveSpaceSummary(String spaceId);
+  Mono<GetSpaceSummaryResponse> retrieveSpaceSummary(String spaceId);
+  
+  Mono<ListDomainsResponse> retrieveDomains();
 	
 	void reset();
 }
--- a/src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessor.java
+++ b/src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessor.java
@@ -6,8 +6,7 @@ import org.cloudfoundry.client.v2.organizations.ListOrganizationsResponse;
 import org.cloudfoundry.client.v2.spaces.GetSpaceSummaryResponse;
 import org.cloudfoundry.client.v2.spaces.ListSpacesResponse;
 import org.cloudfoundry.client.v3.applications.ListApplicationRoutesResponse;
-import org.cloudfoundry.client.v3.domains.ListDomainsResponse;
-import org.cloudfoundry.client.v3.routes.ListRoutesResponse;
+import org.cloudfoundry.client.v3.domains.GetDomainResponse;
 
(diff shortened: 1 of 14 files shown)


src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
```diff
@@ -377,6 +379,13 @@ private void setupPaginatedRequestFetcher() {
...
 				request, r -> this.cloudFoundryClient.spaces().getSummary(r), this.requestTimeoutAppSummary);
 
+  }
+  
+	@Override
+	public Mono<ListDomainsResponse> retrieveDomains() {
+		ListDomainsRequest request = ListDomainsRequest.builder().build();
+		
+		return this.paginatedRequestFetcher.performGenericRetrieval(RequestType.DOMAINS, "(empty)", 
+				request, r -> this.cloudFoundryClient.domainsV3().list(request), this.requestTimeoutAppSummary);
```
eaglerainbow, line 389: You are mixing V2 and V3 interfaces here.
At ours, we still must assume that we only have v2 API endpoints available.

src/main/java/org/cloudfoundry/promregator/scanner/Instance.java
```diff
@@ -8,12 +8,21 @@
 public class Instance {
 	private ResolvedTarget target;
 	private String instanceId;
-	private String accessUrl;
+  private String accessUrl;
+  private boolean isInternal;
```
eaglerainbow, line 12: beans of type boolean do not have the prefix `is`.
```suggestion
  private boolean internal;
```

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessor.java
```diff
@@ -21,7 +22,9 @@
 
 	Mono<ListApplicationsResponse> retrieveAllApplicationIdsInSpace(String orgId, String spaceId);
 
-	Mono<GetSpaceSummaryResponse> retrieveSpaceSummary(String spaceId);
+  Mono<GetSpaceSummaryResponse> retrieveSpaceSummary(String spaceId);
```
eaglerainbow, line 25: Looks like the space/tab settings in your editor are misaligned.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -116,6 +133,18 @@ public String getAccessURL() {
 		 */
 		public void setAccessURL(String accessURL) {
 			this.accessURL = accessURL;
+    }
+    /**
+		 * @return the isInternal
+		 */
+		public boolean getIsInternal() {
```
eaglerainbow, line 140: getters of beans of type boolean do not have the `get` prefix, but always are only `is` prefixed.

eaglerainbow (commented):
I'll add further comments in #190

src/main/java/org/cloudfoundry/promregator/scanner/Instance.java
ScottGuymer: fixed

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
ScottGuymer: The domain api for v2 is marked as deprecated and causes compiler warnings. I dont think the rest of the v2 API does this however and im not sure why.

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessor.java
ScottGuymer: I have added an editorconfig file to make the definition explicit.

ScottGuymer:
closing this in favour of #195 which comes from a different repo.

PR #195, opened 2021-01-27
Internal domains

ScottGuymer:
This is a Work in Progress to show my approach and get feedback.
This replaces #194 and addresses #190 

## Current Situation

Promregator does not currently support routing for internal domains. This PR intends to add support for detecting an internal domain and providing the correct internal urls. See #190 for more info

## Problem Statement
I would like to use promregator with internal routes to keep metrics scraping private and use alternative ports on my apps for metrics servers.

## Solution Approach

- Add retrieval of the available domains from CF
- when detecting the URL for a target match it to a domain
- if the domain is internal use the internal url scheme
- use the port defined in the route for the scraping port

i looks like in the api when you retrieve routes for an app you can see the port that has been mapped there. Even though you have to explicitly use the port in the URL when using an internal route.


ScottGuymer:
Im not sure why the build has broken as i cant see the results..

eaglerainbow:
> Im not sure why the build has broken as i cant see the results..

No surprise - you simply can't: The server is located in my intranet.
I will check the details later - first I want to have a look at the coding ;)

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -192,23 +236,81 @@ public void setNumberOfInstances(int numberOfInstances) {
...
+				// Set the access url to the selected route (without any protocol or path yet)
+				v.setAccessURL(this.determineApplicationRoute(urls, v.getTarget().getOriginalTarget().getPreferredRouteRegexPatterns()));				
+      }      
+           
+			v.setNumberOfInstances(sas.getInstances());
+			
+			return Mono.just(v);
+    });
+        
+    Flux<List<RouteResource>> routeApplicationFlux = osaVectorSpaceFlux.flatMapSequential(v -> this.getAppRoutes(v.getApplicationId()));
```
eaglerainbow, line 248: Performance impact: for each application we will trigger a new request here. If we have thousands of applications (which we already saw in an issue!) this will cause a lot of traffic at the CFCC (even if we cache it properly). 
Moreover, as of now, this isn't guarded yet: All consumers (if they like it or not) would hit this thing here. => not a good option...

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -192,23 +236,81 @@ public void setNumberOfInstances(int numberOfInstances) {
...
+      .findFirst()
+			.orElse(null);
+
+			if (metricsRoute == null) {
+				return Mono.just(v);
+			}
+
+			Destination dest = metricsRoute.getDestinations()
+			.stream()
+			.filter(d -> d.getApplication().getApplicationId().equals(v.getApplicationId()))
```
eaglerainbow, line 270: After several minutes of thinking, I finally understood why you need to do this here. However, that tells me that it might be worth a comment why you do it (one route still can have multiple destinations - and we need to find the right destination...)

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -192,23 +236,81 @@ public void setNumberOfInstances(int numberOfInstances) {
...
+			.orElse(null);
+						
+			if (dest != null) {
+				v.setInternalPort(dest.getPort());
+				v.setRouteId(metricsRoute.getId());
+				v.setDomainId(metricsRoute.getRelationships().getDomain().getData().getId());					
+			}
+            
+      return Mono.just(v);
+    });
```
eaglerainbow, line 281: minor: Indentation

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -192,23 +236,81 @@ public void setNumberOfInstances(int numberOfInstances) {
...
+    
+		Flux<GetDomainResponse> domainFlux = osaVectorApplicationRouteFlux.flatMapSequential(v -> this.cfAccessor.retrieveDomain(v.getDomainId()));
```
eaglerainbow, line 283: What's this those cases where we did not derive a domain yet, e.g. we have a "mixed configuration situation"? Do we want to query `retrieveDomain` for them as well?

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -192,23 +236,81 @@ public void setNumberOfInstances(int numberOfInstances) {
...
+		Flux<OSAVector> osaVectorApplicationRouteDomainFlux = Flux.zip(osaVectorApplicationRouteFlux, domainFlux).flatMap(tuple -> {
+			OSAVector v = tuple.getT1();
+			
+			if (INVALID_DOMAIN == tuple.getT2()) {
```
eaglerainbow, line 287: see above

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
eaglerainbow, line 283: Performance impact: another roundtrip to the CFCC for each application. 
Moreover, the second roundtrip depends on the result of the first (added) roundtrip. 

Performance impact on this case might be significantly less, though, as our cache should have a rather high hit-ratio...

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -192,23 +236,81 @@ public void setNumberOfInstances(int numberOfInstances) {
...
-		
-		Flux<Instance> instancesFlux = osaVectorApplicationFlux.flatMapSequential(v -> {
+
+		Flux<Instance> instancesFlux = osaVectorApplicationRouteDomainFlux.flatMapSequential(v -> {
 			List<Instance> instances = new ArrayList<>(v.getNumberOfInstances());
-			for (int i = 0; i<v.numberOfInstances; i++) {
-				Instance inst = new Instance(v.getTarget(), String.format("%s:%d", v.getApplicationId(), i), v.getAccessURL());
+			for (int i = 0; i<v.numberOfInstances; i++) {        
+        Instance inst = new Instance(v.getTarget(), String.format("%s:%d", v.getApplicationId(), i), v.getAccessURL());
+        inst.setInternal(v.isInternal);
```
eaglerainbow, line 306: minor: indentation

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -296,11 +417,23 @@ public void setNumberOfInstances(int numberOfInstances) {
...
 				return Mono.just(INVALID_SUMMARY);
 			});
+  }
+  
+  private Mono<List<RouteResource>> getAppRoutes(String appId) {
+		return this.cfAccessor.retrieveAppRoutes(appId)
+			.flatMap(response -> {
+				List<RouteResource> routes = response.getResources();
+				if (routes == null) {
+					return Mono.just(INVALID_ROUTES);
```
eaglerainbow, line 427: Hard to understand here, as `INVALID_ROUTES` is null

eaglerainbow (commented):
I have gone through your (productive) coding once (i.e. I haven't had a look at the unit tests yet). The main problem that I see is that we are introducing new roundtrips to the CFCC for each application we have discovered. Based on past experience, this is very costly and even may take down the entire CFCC due to overload. Your calls are even necessary for the legacy/classical case, where no internal domains are supported.
Moreover, Promregator still is commited to provide downward-compatibility to v2-only CloudFoundry platforms. You are using two v3 endpoints, i.e. users of promregator, whose platform still only supports v2 endpoints, will receive a regression. That means the current state is still incompatible. I see two options to get out of that hell:
* Upon startup we check which version of the platform is running - and by that see if the v3 endpoints are available. Only if they are, we "enable" internal domains. (preferred solution)
* We introduce a manual configuration option, where the user may declare that its CF platform has v3 support and by that toggles on your additional logic.

One more thing which I came across: You are adding additional (cache timing) configuration parameters. We need to explain them in the config.md documentation...

eaglerainbow:
> I will check the details later 

The job failed due to an infrastructural problem (nothing you could fix so far). I will try to get the job fixed (don't know yet how), but it's unclear to me how much benefit that will bring at the current state. I think we should focus more on the code-review comments first.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: Yes I think the domain trip should be very well cached. I can see how the application route round trip could add some overhead.

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeine.java
```diff
@@ -45,6 +49,12 @@
 	
 	@Value("${cf.cache.timeout.application:300}")
 	private int refreshCacheApplicationLevelInSeconds;
+
+	@Value("${cf.cache.timeout.domain:300}")
```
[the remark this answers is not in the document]
ScottGuymer: Agreed I'll set it to 3600 as its unlikely to change more frequently than hour by hour.

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheClassic.java
```diff
@@ -399,4 +410,23 @@ public void reset() {
...
+	public Mono<ListApplicationRoutesResponse> retrieveAppRoutes(String appId) {
+		    /*
+		 * not implemented cache here 
+		 */
+		return this.parent.retrieveAppRoutes(appId);
+	}
+
+	@Override
+	public void invalidateCacheDomain() {
+		// TODO Auto-generated method stub
```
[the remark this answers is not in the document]
ScottGuymer: removed

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheClassic.java
```diff
@@ -399,4 +410,23 @@ public void reset() {
...
+		// Nothing to do here as not cached
+	}
+
+	@Override
+	public void invalidateCacheRoutes() {
+		// TODO Auto-generated method stub
```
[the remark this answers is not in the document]
ScottGuymer: removed

.editorconfig
```diff
@@ -0,0 +1,11 @@
...
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.java]
+indent_style = tab
+indent_size = 2
```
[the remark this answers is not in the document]
ScottGuymer: Updated to 4

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
```diff
@@ -85,7 +89,13 @@
...
 	
 	@Value("${cf.request.timeout.appSummary:4000}")
-	private int requestTimeoutAppSummary;
+  private int requestTimeoutAppSummary;
+  
+  @Value("${cf.request.timeout.domains:4000}")
+  private int requestTimeoutDomains;
+
+  @Value("${cf.request.timeout.appRoutes:4000}")
+	private int requestTimeoutAppRoutes;
```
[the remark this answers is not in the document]
ScottGuymer: Strange that these all looked right in my editor until i changed the editorconfig spacing to 4. Now i see them :)

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: Do you mean in the case when where we cant find a matching route for the selected url?

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: fixed

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeine.java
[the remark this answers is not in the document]
ScottGuymer: Does the `cf.cache.expiry.domain` need changing too?

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorSimulator.java
```diff
@@ -174,5 +207,47 @@ public void reset() {
...
 	}
 
+	@Override
+	public Mono<ListApplicationRoutesResponse> retrieveAppRoutes(String appId) {
+		List<RouteResource> routes = new LinkedList<>();
+
+		String instanceId = appId.replace(APP_UUID_PREFIX, "");
+
+		RouteResource res = RouteResource.builder()
+      .id("id")
```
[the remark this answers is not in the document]
ScottGuymer: fixed

src/main/java/org/cloudfoundry/promregator/cfaccessor/RequestType.java
```diff
@@ -6,7 +6,9 @@
 	SPACE("space", "retrieveSpaceId"),
 	SPACE_IN_ORG("space", "retrieveAllSpaceIdsInOrg"),
 	ALL_APPS_IN_SPACE("allApps", "retrieveAllApplicationIdsInSpace"),
-	SPACE_SUMMARY("spaceSummary", "retrieveSpaceSummary"),
+  SPACE_SUMMARY("spaceSummary", "retrieveSpaceSummary"),
+  DOMAINS("domains", "retrieveDomains"),
+  APP_ROUTES("appRoutes", "retrieveAppRoutes"),
```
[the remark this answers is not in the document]
ScottGuymer: done

src/main/java/org/cloudfoundry/promregator/scanner/Instance.java
```diff
@@ -8,12 +8,21 @@
 public class Instance {
 	private ResolvedTarget target;
 	private String instanceId;
-	private String accessUrl;
+  private String accessUrl;
+  private boolean internal;
```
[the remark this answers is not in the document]
ScottGuymer: Added what i think is needed.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -30,7 +36,11 @@
 	private static final Logger log = LoggerFactory.getLogger(ReactiveAppInstanceScanner.class);
 	private static final String INVALID_ORG_ID = "***invalid***";
 	private static final String INVALID_SPACE_ID = "***invalid***";
-	private static final Map<String, SpaceApplicationSummary> INVALID_SUMMARY = new HashMap<>();
+  private static final Map<String, SpaceApplicationSummary> INVALID_SUMMARY = new HashMap<>();
+  private static final List<RouteResource> INVALID_ROUTES = null;
+  private static final GetDomainResponse INVALID_DOMAIN = null;
+
+  private static final int DEFAULT_INTERNAL_METRICS_PORT = 9090;
```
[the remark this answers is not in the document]
ScottGuymer: I just removed this as it won't happen because the port comes from the API.
If it is needed it can be added later. This PR is big enough :D

.editorconfig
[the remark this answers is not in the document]
ScottGuymer: Before i changed this none of the indentation problems you caught were visible to me.. Live and learn

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: Added a clarifying comment

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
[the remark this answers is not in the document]
ScottGuymer: I am not familiar with reactor at all so I kind of assumed that they would need to be in order. I can see that changing it to use `initialOSAVectorFlux` and running the tests again that it seems to work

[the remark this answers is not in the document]
ScottGuymer: That should work a treat...

ScottGuymer: There is an API the returns _all_ routes for a space. This would result in a larger response but many less requests. We could then cache that response.

Its available in V2 api too so that would just leave the domain part in V3.

Sound like a good refactor?

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -192,23 +236,81 @@ public void setNumberOfInstances(int numberOfInstances) {
...
+			v.setNumberOfInstances(sas.getInstances());
+			
+			return Mono.just(v);
+    });
+        
+    Flux<List<RouteResource>> routeApplicationFlux = osaVectorSpaceFlux.flatMapSequential(v -> this.getAppRoutes(v.getApplicationId()));
+		Flux<OSAVector> osaVectorApplicationRouteFlux = Flux.zip(osaVectorApplicationFlux, routeApplicationFlux).flatMap(tuple -> {
+			OSAVector v = tuple.getT1();
+			
+			if (INVALID_ROUTES == tuple.getT2()) {
```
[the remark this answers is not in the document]
ScottGuymer: I think this is my lack of familiarity again. I changed it to check for an empty list rather than a null list.

If the list is empty (ie there are no routes) then the target will be dropped.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: I removed this as im now not sure it is required. My understanding is that flux/mono wont allow there to be a null value here anyway and if there is an error with the API call it will have thrown way before this.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: I can again return a list of all domains for a particular space. This would reduce the chatter back and for for individual applications.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: I changed this to return an empty list rather than null. I think that makes more sense. 

We can then decide whether to drop that target or not in the flux rather than getting an error.

ScottGuymer:
I have processed most of the comments now.

I understand the issues with the V3 vs V2 as well as the performance issues.

What I propose is

- Query all routes for a space rather than per application. We can cache this and filter as required.
- Query all domains for a space and filter for the right one based on the route

I have had a look at using the v2 API for this. Originally i didnt think it was possible. But i ahev looked deeped and while I think its feasible i fear it introduces yet more round trips but they are on the account level so they could be cached.

I can try and refactor to use the v2 api as follows
 - match the preferred URL to the route object from all routes in the space
 - Match the the route_mapping for that route/app to get the port number
 - get the domain from the list of all domains to determine if it is internal or not.

All of these calls can be made for the space/login so can be single calls and cached rather than per app.

WDYT?

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -192,23 +230,80 @@ public void setNumberOfInstances(int numberOfInstances) {
...
+			// a route can have many different destinations
+			// once we get this we can use the routeID and the domainID
+			Destination dest = metricsRoute.getDestinations()
+				.stream()
+				.filter(d -> d.getApplication().getApplicationId().equals(v.getApplicationId()))
+				.findFirst()
+				.orElse(null);
+						
+			if (dest != null) {
+				v.setInternalPort(dest.getPort());
```
eaglerainbow, line 273: BTW: After having a read a little about container-to-container communication in CF, I understand that this is very implicit here. You assume that the application gets another route which is set like this:

``` 
cf map-route myapp app.internal --hostname myapphost --port 9090 
```

I think the scenario we are having here needs some further documentation as well, such that users will be able to configure their application properly.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -192,23 +230,80 @@ public void setNumberOfInstances(int numberOfInstances) {
 
 			List<String> urls = sas.getUrls();
 			if (urls != null && !urls.isEmpty()) {
-				v.setAccessURL(this.determineAccessURL(v.getTarget().getProtocol(), urls, v.getTarget().getOriginalTarget().getPreferredRouteRegexPatterns(), v.getTarget().getPath()));
-			}
+				// Set the access url to the selected route (without any protocol or path yet)
+				v.setAccessURL(this.determineApplicationRoute(urls, v.getTarget().getOriginalTarget().getPreferredRouteRegexPatterns()));
```
eaglerainbow, line 234: Here is another implicit assumption: Let us assume that promregator was deployed as CF app "promregator", then a network policy was required as well:

``` 
cf add-network-policy promregator --destination-app apptoscrape --protocol tcp --port 9090
``` 

==> another thing for documentation

Additionally, the route which has also to be mapped (see other comment below) also needs to be entered in the `preferredRouteRegex` of that application.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
eaglerainbow: yes, for instance... Let's say, someone has configured a single route pattern which has an error and does not match against any route...
In your case, you would feed `retrieveDomain` with `null`.

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
```diff
@@ -377,6 +387,21 @@ private void setupPaginatedRequestFetcher() {
...
 				request, r -> this.cloudFoundryClient.spaces().getSummary(r), this.requestTimeoutAppSummary);
 
+  }
+  
+	@Override
+	public Mono<GetDomainResponse> retrieveDomain(String domainId) {       
+    GetDomainRequest request = GetDomainRequest.builder().domainId(domainId).build();
+		
+		return this.paginatedRequestFetcher.performGenericRetrieval(RequestType.DOMAINS, "(empty)", 
+				request, r -> this.cloudFoundryClient.domainsV3().get(request), this.requestTimeoutDomains);
```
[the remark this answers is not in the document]
eaglerainbow: Just an idea: I found https://apidocs.cloudfoundry.org/15.3.0/shared_domains/list_all_shared_domains.html and https://apidocs.cloudfoundry.org/15.3.0/shared_domains/retrieve_a_particular_shared_domain.html

What about the idea to retrieve all domains in one shot? The number of domains on a cloud foundry deployment should not be skyrocketing...

eaglerainbow:
> I can try and refactor to use the v2 api as follows

>     match the preferred URL to the route object from all routes in the space
>     Match the the route_mapping for that route/app to get the port number
>     get the domain from the list of all domains to determine if it is internal or not.

> All of these calls can be made for the space/login so can be single calls and cached rather than per app.

> WDYT?

Far much better... This means though that you will need (initially) three roundtrips in total, if I am not mistaken. However, in a scenario where you have hundreds (or even thousands) of apps being scraped, that could make a whole lot of a difference. The unanswered question to me is: can we get this into this reactive coding logic well enough to still be maintainable :-|

@trac

eaglerainbow:
@tlkamp Hi Traci, we are screwing here on the route derivation logic. We might end up with a new load of additional round trips to the CF Cloud Controller. The idea is to get a new feature in, such that you may also scrape using container-to-container communication (through an/the internal domain). This could improve security as the scraping endpoints could be registered on a separate port, which is only exposed internally. 
How much would you think does this hurt you in your "high-load scenario"? Do you need further details to assess?

tlkamp:
@eaglerainbow Hi!!

I would need some more information to understand how this might impact the (still) undersized infrastructure I am working with. How often do the extra round trips occur? On startup and then again when the cache expires? Does the internal domain routing work if I am not hosting Promregator on Cloud Foundry?

I would not mind trying a pre-release/alpha/beta version of this functionality and coming back with results. I should be able to quantify the existing number of round trips and then compare that with the new feature.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: Yes I think this would need to be documented and might be where I need to allow for a default port at the target level for cases where the route port is null.

For now I will leave this as a TODO so that it can be implemented.
I will add test cases for it too.

I think the right solution is to add a default port at the target level so it can fall back to something configured by the suer.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: Yes the network policies are important and need to be created to allow promregator to speak to anything internally. This will have to go into the docs.

I will start an additional file on internal routing.

ScottGuymer:
I agree with @eaglerainbow at the moment what i have generates way too much traffic. 

Let me refactor it to use larger scoped queries that will need to happen less often, eg getting all routes for a space rather than each app.

eaglerainbow:
> Let me refactor it to use larger scoped queries that will need to happen less often, eg getting all routes for a space rather than each app.

@ScottGuymer Agreed.

@tlkamp Let Scott first do the refactoring. Once he's finished, I will (somehow) bundle a test version for you, which you will get together with some more details on how to see the roundtrips. Then let's test in your environment what it means...

ScottGuymer:
@eaglerainbow  @tlkamp 

I have now finished the bulk of the refactor. Still needs some work but is functional and should give you an idea of what I am trying to achieve.

We are now
- using the v2 api
- gathering routes, route_mappings and domains all at once rather than per app

Let me know what you think and what i need to improve

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorSimulator.java
```diff
@@ -174,5 +186,81 @@ public void reset() {
...
+			RouteResource res = RouteResource.builder()
+			.metadata(
+				Metadata.builder()
+				.id(APP_ROUTE_UUID+i)
+				.createdAt(CREATED_AT_TIMESTAMP)
+				.build())
+			.entity(
+				RouteEntity.builder()
+				.host(APP_HOST_PREFIX+i)			
+				.path("path")
```
eaglerainbow, line 231: using `APP_ROUTE_PATH` here?

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
```diff
@@ -87,6 +93,12 @@
 	@Value("${cf.request.timeout.appSummary:4000}")
 	private int requestTimeoutAppSummary;
 	
+	@Value("${cf.request.timeout.domains:4000}")
```
eaglerainbow, line 96: Documentation of new configuration parameter needed (later on)

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
```diff
@@ -87,6 +93,12 @@
...
+	private int requestTimeoutDomains;
+
+	@Value("${cf.request.timeout.appRoutes:4000}")
```
eaglerainbow, line 99: Documentation of new configuration parameter needed (later on)

src/main/java/org/cloudfoundry/promregator/scanner/AppRouteVector.java
```diff
@@ -0,0 +1,67 @@
...
+		return port;
+	}
+	public void setPort(int port) {
+		this.port = port;
+	}
+	public void setAppId(String appId) {
+		this.appId = appId;
+	}
+	public String getUrl() {
+		// TODO check neither are null
```
eaglerainbow, line 63: Open todo?

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -50,103 +57,228 @@
...
+	private Flux<AppRouteVector> determineRoutes(String orgId, String spaceId) {
+		Mono<ListRouteMappingsResponse> routeMappings = this.cfAccessor.retrieveRouteMappings();
+
+		Flux<RouteMappingResource> something = routeMappings
+				.flatMapMany(mapper -> Flux.fromIterable(mapper.getResources()));
+
+		Flux<AppRouteVector> routeMappingFlux = something.map(mapper -> {
+			AppRouteVector route = new AppRouteVector();
+
+			// TODO some validation here
```
eaglerainbow, line 204: Open todo?

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
```diff
@@ -377,6 +389,30 @@ private void setupPaginatedRequestFetcher() {
...
+		return this.paginatedRequestFetcher.performGenericRetrieval(RequestType.APP_ROUTES, spaceId, 
+		request, r -> this.cloudFoundryClient.spaces().listRoutes(request), this.requestTimeoutAppRoutes);
+	}
+
+	@Override
+	public Mono<ListRouteMappingsResponse> retrieveRouteMappings() {
+		ListRouteMappingsRequest request = ListRouteMappingsRequest.builder().build();
+
+		return this.paginatedRequestFetcher.performGenericRetrieval(RequestType.APP_ROUTES, "(empty)", 
+		request, r -> this.cloudFoundryClient.routeMappings().list(request), this.requestTimeoutAppRoutes);
```
eaglerainbow, line 416: Wow. The response here might be very big, as we don't have any filtering here.
Even worse, it means that we will read all route mappings of all application on the entire installation. So, if we want to scrape 2 apps out of 10000 apps, then we still will have to store all 10000*n route mappings here.

Did you consider/estimate the memory consumption this could mean for us?

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -50,103 +57,228 @@
...
 		}
-		
+
 	}
-	
+
+	private Flux<AppRouteVector> determineRoutes(String orgId, String spaceId) {
+		Mono<ListRouteMappingsResponse> routeMappings = this.cfAccessor.retrieveRouteMappings();
+
+		Flux<RouteMappingResource> something = routeMappings
```
eaglerainbow, line 198: minor: `something` is a strange naming: Perhaps `routeMappingResourceFlux` is better

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -50,103 +57,228 @@
...
+				.flatMapMany(mapper -> Flux.fromIterable(mapper.getResources()));
+
+		Flux<AppRouteVector> routeMappingFlux = something.map(mapper -> {
+			AppRouteVector route = new AppRouteVector();
```
eaglerainbow, line 202: Consider adjusting constructor here?

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -50,103 +57,228 @@
...
+			route.setAppId(mapper.getEntity().getApplicationId());
+			route.setPort(mapper.getEntity().getApplicationPort());
+			route.setRouteId(mapper.getEntity().getRouteId());
+			return route;
+		});
+
+		Flux<List<RouteResource>> spaceRoutes = routeMappingFlux.flatMapSequential(v -> {
+			return this.cfAccessor.retrieveSpaceRoutes(spaceId).flatMap(mapper -> Mono.just(mapper.getResources()));
+		});
+		Flux<AppRouteVector> routeInstanceFlux = Flux.zip(routeMappingFlux, spaceRoutes).flatMap(tuple -> {
```
eaglerainbow, line 215: Looks a little dangerous here: What if the CFCC just had a small hiccup before with one of the spaces when reading the space's routes (e.g. returning an empty `getResources()`)?
Or there was some latency happening and the configuration has just been changed....

At least, consistency should be checked afterwards.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -50,103 +57,228 @@
...
+			if (routes.size() == 0) {
+				// NB: This drops the current instance route!
+				return Mono.empty();
+			}
+
+			// find the route to match this InstanceRoute
+			RouteResource route = routes.stream().filter(r -> r.getMetadata().getId().equals(v.getRouteId()))
+					.findFirst().get();
+
+			// TODO what if we dont find???
```
eaglerainbow, line 228: yeah, that's what I mean ;-)

eaglerainbow (commented):
Just some first raw statements that I came across when running through it the first time. 
In particular when looking at https://apidocs.cloudfoundry.org/15.3.0/routes_mapping/list_all_route_mappings.html, I see that the idea of using v2 api only can make things very tricky. 
Also this "reading of all Route Mappings" may be much better for the CFCC from a load perspective, but it will 
a) triggers a memory consumption discussion, and 
b) the cache timing will become interesting, because as soon as one single application (to scrape) changes its routes, the cache should be invalidated. Given that this is rather common when doing blue-green deployment, the cache TTL will be rather short and we will have to ask for all the route mapping metadata over and over again. 

I have to sleep another night over this. Perhaps I have to even need to wrap my head around this a little more on the weekend...

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
ScottGuymer: As far as I can see this is the only way to do this with v2 without querying each app individually. There is no space or org level API for route_mappings.

It will _only_ be the route mappings you have the permissions to see (if that helps).

I didnt do much work on checking how this scales at load yet. However the route_mappings are quite small per mapping so im not sure even 10,000 would be _too_ challenging

Each route mapping is something like this

```
{
 "metadata": {
    "guid": "589bce61-48d3-476a-b90a-xxxxxxxxxxx",
    "url": "/v2/route_mappings/589bce61-48d3-476a-b90a-xxxxxxxxxxx",
    "created_at": "2021-02-01T12:16:24Z",
    "updated_at": "2021-02-01T12:16:24Z"
 },
 "entity": {
    "app_port": 80,
    "app_guid": "a802b42a-adb9-47c8-914c-xxxxxxxxxxx",
    "route_guid": "6b8ebb34-8ae1-4d1f-92a1-xxxxxxxxxxx",
    "app_url": "/v2/apps/a802b42a-adb9-47c8-914c-xxxxxxxxxxx",
    "route_url": "/v2/routes/6b8ebb34-8ae1-4d1f-92a1-xxxxxxxxxxx"
 }
```

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: :D it was late and i had run out of inspiration

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: Add the params to the constructor rather than setting them later?

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: Yup i need to go back and add in some "robustness" once we are happy with the implementation..

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: For sure. I focussed on getting this "working" for now. to prove its possible before going too much furtehr.

ScottGuymer:
Thanks for taking a look @eaglerainbow.

It took me a while to get back round to it again. 

Using the v2 API does make it a lot tricker than it might be with v3, especially when trying not to query each app individually.

Im not sure there is any real other way around though...

eaglerainbow:
Wrapping my head around it: Could you have a look at the example mentioned at https://apidocs.cloudfoundry.org/15.3.0/spaces/get_space_summary.html ? I mean there in particular the section `apps[].routes[].domain`. We [are already retrieving](https://github.com/promregator/promregator/blob/976e35ed0b070c8b3a8c815c9652dd74089f648d/src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java#L305) that information anyway today. Perhaps this could allow us applying the following approach:

1. Retrieve all domains (as you already do). Generate a `HashSet` out of those, which are internal. In case of error/fallback, we take the emptySet.
2. Look at the application in the space summary. There you may determine which routes are used. If one of these routes uses a domain which is in the set of (1), we prefer that one (and know that it is to be scraped internally).
3. The port to scrape against is mentioned in  `apps[].routes[].port`.

Does this make sense to you?
Or is there a certain reason why you had dropped that approach (and I just didn't notice that yet)?

ScottGuymer:
I think that would be perfect. However, the v2 API seems to have different ports in mind... The port exposed on the route's entity is not the same value as the port on the route mapping entity... The port isnt even always included on the route entity for some reason. I dont know enough about the API here to be sure what is happening.

Maybe i need to explicitly define a port in the route entity for it to work here...

A concrete example from my setup.

```
❯ cf curl /v2/routes/1e587cf6-fcaf-43bf-857c-xxxxxxxxxx
{
   "metadata": {
      "guid": "1e587cf6-fcaf-43bf-857c-xxxxxxxxxx",
      "url": "/v2/routes/1e587cf6-fcaf-43bf-857c-xxxxxxxxxx",
      "created_at": "2021-01-29T09:03:24Z",
      "updated_at": "2021-01-29T09:03:24Z"
   },
   "entity": {
      "host": "frontend",
      "path": "",
      "domain_guid": "409ec4df-d54d-4a93-8428-xxxxxxxxxx",
      "space_guid": "e51e69a1-4cc7-4601-a48e-xxxxxxxxxx",
      "service_instance_guid": null,
      "port": null,
      "domain_url": "/v2/shared_domains/409ec4df-d54d-4a93-8428-xxxxxxxxxx",
      "space_url": "/v2/spaces/e51e69a1-4cc7-4601-a48e-xxxxxxxxxx",
      "apps_url": "/v2/routes/1e587cf6-fcaf-43bf-857c-xxxxxxxxxx/apps",
      "route_mappings_url": "/v2/routes/1e587cf6-fcaf-43bf-857c-xxxxxxxxxx/route_mappings"
   }
}
❯ cf curl /v2/routes/1e587cf6-fcaf-43bf-857c-xxxxxxxxxx/route_mappings
{
   "total_results": 1,
   "total_pages": 1,
   "prev_url": null,
   "next_url": null,
   "resources": [
      {
         "metadata": {
            "guid": "2f95b1cd-404f-465e-881a-xxxxxxxxxx",
            "url": "/v2/route_mappings/2f95b1cd-404f-465e-881a-xxxxxxxxxx",
            "created_at": "2021-01-29T09:03:43Z",
            "updated_at": "2021-01-29T09:03:43Z"
         },
         "entity": {
            "app_port": 8080,
            "app_guid": "0e7c022b-de03-4d2a-afa2-xxxxxxxxxx",
            "route_guid": "1e587cf6-fcaf-43bf-857c-xxxxxxxxxx",
            "app_url": "/v2/apps/0e7c022b-de03-4d2a-afa2-xxxxxxxxxx",
            "route_url": "/v2/routes/1e587cf6-fcaf-43bf-857c-xxxxxxxxxx"
         }
      }
   ]
}
```

ScottGuymer:
On some further testing this does not seem to be consistent _at all_ across the v2/v3 api and even different usages for apps...

eaglerainbow:
That's too bad... 😢 I thought I'd found an elegant solution to all our problems. Unfortunately, it calls for even more problems. 
In essence it boils down to the two options:
* Go for v2 approach (mainly the current state of this PR here)
* Recheck for a v3 approach (if we can achieve a better performance behavior there) and toggle it. 

I'll wrap my head around option 2 for a while, but can only start doing so starting tomorrow.

ScottGuymer:
I have done a little more of a deep dive into what the port shows in the API.

- v2 Routes - shows the TCP port you specify when you create the route
- v2 Route Mappings - im not actually sure what this is showing even when you specify a TCP port when creating a mapping it sometimes shows null. In other routes it shows the port mapped to the container (docker in my case) but this seems to be an automatic thing...
- v3 Routes (destinations) shows the port used to speak to the container (again docker) but im not sure if this is only automatic or something that can be defined when you deploy using the API.

All of this leads me to thinking that the "automatic" configuration of the port in use from the API is not going to work. Especially not in my configuration where i actually expose 2 ports from the container (1 for the app and one for the metrics).

So what i propose is.

- Add in a target config value for `internalRoutePort` to specify the port to use for scaping that target only if its an internal route
- Add in a global config for a "default" `internalRoutePort` as a fallback
- Use the all domains query to cache domains
- Remove all the other calls to routes and mapping etc and just use the domain in the space summary to connect to the domains retrieved below and use the configured port for scraping.


This should solve the performance issues, reduce complexity and rule out any odd stuff with ports in the API.

ScottGuymer:
So I have refactored (hopefully for the last time :D) and pushed the updates.

I went a slightly different route here and didn't switch on or off the internal routing based on the setting of `internalRoutePort`.

My reasoning being,  if have promregator configured on an internal route then the behaviour would at best be flawed and most likely just plain broken (hence why I am here with this PR :D ).

Current best case for this working is the following circumstances
- you are running promregator inside CF
- your app exposes metrics on port 80
- you have the required network policies in place

At best here the scape is hitting _an_ instance based on DNS RR, but not specific ones and not all of them. So any metrics you were getting here are probably not that accurate.

So if someone has this configured today and its _working_ , then it isn't really and they with this change they might be prompted to go investigate it.

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeine.java
```diff
@@ -45,6 +52,12 @@
 	
 	@Value("${cf.cache.timeout.application:300}")
 	private int refreshCacheApplicationLevelInSeconds;
+
+	@Value("${cf.cache.timeout.domain:3600}")
```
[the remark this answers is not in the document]
ScottGuymer: Added

ScottGuymer:
Its showing an issue bullding the code but i cant navigate to the results to see the issue..

eaglerainbow:
Sorry for my delay -- too much other stuff around me in the last 2 weeks.

> Its showing an issue bullding the code but i cant navigate to the results to see the issue..

Unfortunately, you can't look into this; I'll have a look and, if there is something for you to do, will give you feedback...

eaglerainbow:
For the sake of completeness, here's what the logs says (and why the voter failed):

```
01:00:24 [ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 3.99 s <<< FAILURE! - in org.cloudfoundry.promregator.cfaccessor.CFAccessorCacheCaffeineTimeoutTest
01:00:24 [ERROR] testRetrieveSpaceSummary  Time elapsed: 0.525 s  <<< FAILURE!
01:00:24 org.mockito.exceptions.verification.TooFewActualInvocations: 
01:00:24 
01:00:24 cFAccessor.retrieveSpaceSummary("dummy");
01:00:24 Wanted 2 times:
01:00:24 -> at org.cloudfoundry.promregator.cfaccessor.CFAccessorCacheCaffeineTimeoutTest.testRetrieveSpaceSummary(CFAccessorCacheCaffeineTimeoutTest.java:130)
01:00:24 But was 1 time:
01:00:24 -> at org.cloudfoundry.promregator.cfaccessor.CFAccessorCacheCaffeine$SpaceSummaryCacheLoader.asyncLoad(CFAccessorCacheCaffeine.java:139)
01:00:24 
01:00:24 	at org.cloudfoundry.promregator.cfaccessor.CFAccessorCacheCaffeineTimeoutTest.testRetrieveSpaceSummary(CFAccessorCacheCaffeineTimeoutTest.java:130)
```

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -168,84 +222,136 @@ public void setNumberOfInstances(int numberOfInstances) {
...
 			
 			v.setNumberOfInstances(sas.getInstances());
-			
+
 			return Mono.just(v);
 		});
-		
+
+		Flux<List<DomainResource>> domainFlux = osaVectorApplicationFlux.flatMapSequential(v -> {
+			return this.cfAccessor.retrieveAllDomains(v.getOrgId()).flatMap(mapper -> Mono.just(mapper.getResources()));
```
eaglerainbow, line 271: might want to do 
```suggestion
			return this.cfAccessor.retrieveAllDomains(v.getOrgId()).map(mapper -> mapper.getResources());
```
instead (mainly just easier to read).

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -263,60 +369,71 @@ public void setNumberOfInstances(int numberOfInstances) {
...
 		String internalPath = path;
 		while (internalPath.startsWith("/")) {
 			internalPath = internalPath.substring(1);
 		}
-		
+
 		return applUrl + internalPath;
 	}
 
+	private String formatInternalAccessURL(final String url, final String path, final int internalRoutePort,
```
eaglerainbow, line 424: I find the name of `url` a little confusing. This is more something like `hostnameDomain` or so...

docs/internal-routing.md
```diff
@@ -0,0 +1,15 @@
+# Internal Routing
+
+It's possible to deploy promregator into cloudfoundry as a cloud foundry app and this enables the possibility of routing all promregator calls internally so that they don't leave your environment.
+
+## Defining an internal target
+
+When promregator resolves your targets it will automatically detect if the route that has been selected is using an internal domain. If an internal domain is detected then your targets will be scraped internally. When this happens promregator uses some slightly different logic to scrape your application. This is because the internal router does not understand the instance header used for regular promregator scraping. When scraping internally it uses a the url format `<instance_number>.<hostname>.<domain>:<port>`. Because of this format a port number is needed for scraping. A default port number of `8080` is defined but you can also define the `internalRoutePort` on a per target basis.
```
eaglerainbow, line 7:
```suggestion
When promregator resolves your targets it will automatically detect if the route that has been selected is using an internal domain. You may use configuration parameter `preferredRouteRegex` to guide Promregator to the appropriate route. If an internal domain is detected, then your targets will be scraped internally. When this happens promregator uses some slightly different logic to scrape your application. This is because the internal router does not understand the instance header used for regular promregator scraping. When scraping internally it uses the url format `<instance_number>.<hostname>.<domain>:<port>`. Because of this format a port number is needed for scraping. A default port number of `8080` is defined (see also configuration parameter `defaultInternalRoutePort`), but you can also define the `internalRoutePort` on a per target basis.
```

eaglerainbow (commented):
I am much happier with the current state of this! So, I think that all the refactoring you did and discussion finally starts to pay off!
In particular, I like the approach of just interpreting the domain's internal flag and then switch over the behavior. The `defaultInternalPort` also seems to do the trick.
My main comments below are mainly going into the direction of documentation.
Unfortunately, I didn't find the time yet to have a look into your unit tests. I still want to do this - and perhaps along then we also may find the issue with the the unit test reported by the voter.

docs/internal-routing.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Network policies
+
+To enable internal communication between containers you must define network policies to allow the traffic between promregator and the target app. When defining the network policy it's important to ensure you use the same port that you expect to scrape on and configure in the `internalRoutePort` value of your target.
+
+```sh
+cf add-network-policy promregator --destination-app apptoscrape --protocol tcp --port 9090
+```
```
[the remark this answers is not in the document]
ScottGuymer: Added

docs/internal-routing.md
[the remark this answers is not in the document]
ScottGuymer: Added

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheClassic.java
```diff
@@ -260,10 +269,73 @@ public void setupMaps() {
 		 */
 
 		return mono;
-	}
+	}		
+
+  	private Mono<GetSpaceSummaryResponse> spaceSummaryCacheLoader(String spaceId) {
+	Mono<GetSpaceSummaryResponse> mono = this.parent.retrieveSpaceSummary(spaceId).cache();
```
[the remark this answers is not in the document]
ScottGuymer: fixed

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorSimulator.java
```diff
@@ -174,5 +185,26 @@ public void reset() {
...
+	public Mono<ListOrganizationDomainsResponse> retrieveAllDomains(String orgId) {		
+		List<DomainResource> domains = new ArrayList<DomainResource>();
+
+		for (int i = 1;i<=100;i++) {
+			
+			DomainResource domain = DomainResource.builder()
+			.entity(
+				DomainEntity.builder()
+				.name(SHARED_DOMAIN)
+				.internal(false)
```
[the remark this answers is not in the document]
ScottGuymer: Added a single internal domain

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
ScottGuymer: Yup i have updated

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
```diff
@@ -377,6 +382,14 @@ private void setupPaginatedRequestFetcher() {
 		return this.paginatedRequestFetcher.performGenericRetrieval(RequestType.SPACE_SUMMARY, spaceId, 
 				request, r -> this.cloudFoundryClient.spaces().getSummary(r), this.requestTimeoutAppSummary);
 
-	}
+  }
+
+	@Override
+	public Mono<ListOrganizationDomainsResponse> retrieveAllDomains(String orgId) {
+
+		ListOrganizationDomainsRequest request = ListOrganizationDomainsRequest.builder().organizationId(orgId).build();
```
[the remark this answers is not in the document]
ScottGuymer: Nothing specific I don't think..

Probably just that it was like this from one of the intermediate refactors.

I'm not sure if there is any concrete difference between a global domain query and an org one...

The advantage I can see here is that the org is only a small subset and on a large install the global one could be a considerably larger list..

Happy to change it but I cant see it being a substantial difference either way.

ScottGuymer:
I am also much happier with this.

All the tests pass for me locally and I haven't seen any issues with that specific test. (it's not even testing any code that I have changed). I am hoping it's just a transient error. 

The only tests I could think about would be to test if there were no matching domain entities returned for an app. However, i cant see how a CF instance could get into this state. It is guarded against in the code and can be tested by commenting out the  `UNITTEST_ADDITIONAL_SHARED_DOMAIN` entity in `CFAccessorMock.java`. When you comment this out all of the app instance scanner tests still pass. Maybe this is sufficient.

ScottGuymer:
@eaglerainbow im keen to get this complete now as its becoming something of an impediment for me.

What are the next steps?

eaglerainbow:
Having a day off today, it has been my plan to check on this one (you are at position 2 on my agenda today :) )

src/test/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorSimulatorTest.java
```diff
@@ -2,6 +2,8 @@
 
 import java.util.List;
 
+import org.cloudfoundry.client.v2.domains.DomainResource;
+import org.cloudfoundry.client.v2.organizations.ListOrganizationDomainsResponse;
```
eaglerainbow, line 6: Looks superfluous to me.

Would be good, though, if we could have a little coverage on the `CFAccessorSimulator` though.

docs/internal-routing.md
```diff
@@ -0,0 +1,19 @@
...
+When promregator resolves your targets it will automatically detect if the route that has been selected is using an internal domain. You may use configuration parameter `preferredRouteRegex` to guide Promregator to the appropriate route. If an internal domain is detected, then your targets will be scraped internally. When this happens promregator uses some slightly different logic to scrape your application. This is because the internal router does not understand the instance header used for regular promregator scraping. When scraping internally it uses the url format `<instance_number>.<hostname>.<domain>:<port>`. Because of this format a port number is needed for scraping. A default port number of `8080` is defined (see also configuration parameter `defaultInternalRoutePort`), but you can also define the `internalRoutePort` on a per target basis.
+
+When using internal routes for scraping you should only have a single route per application. This is because the instance identification is via DNS and not headers it is not possible to distinguish different apps on the same route. If multiple apps are on the same route it is possible that there are inconsistent scraping results.
+
```
eaglerainbow, line 10: Should we also mention that normally the user wants to set `promregator.targets[].protocol` to `http`, if running on internal routing?

eaglerainbow (commented):
No big things anymore  - also in the tests.
Also the unit tests went through on the Jenkins infrastructure. 👍 

There is one more thing which came to my mind: https://github.com/promregator/promregator/blob/12c64bc9b25047eaea478a10d83b725046b94538/src/main/java/org/cloudfoundry/promregator/fetcher/CFMetricsFetcher.java#L125 
Sending this header does not make any sense anymore. The only place where `CFMetricsFetcher` is created is https://github.com/promregator/promregator/blob/5f9aba151858ddcd2f5741b67333f3c0c8201db0/src/main/java/org/cloudfoundry/promregator/endpoint/AbstractMetricsEndpoint.java#L363. I think, the interface of the constructor should be enhanced by a `boolean withRoutingHeader` (or alike) - and should be set depending on the `Instance`'s state of `internal`. This would clean up the outbound request as well. wdyt?

eaglerainbow:
Action plan from my perspective:
- Mop up the rest of the comments here
- Merge to `master`.
- Create temporary docker-based release candidate version for Traci (new minor release, as we are adding a feature).
- Let Traci test that thing (I don't expect too much issues with the latest variant, though). Await and handle feedback.
- (Iterate with release candidate versions, if applicable).
- Deliver new version. (Reminder to myself: Don't forget the patchnotes/changelog).

docs/internal-routing.md
ScottGuymer: indeed added.

src/test/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorSimulatorTest.java
ScottGuymer: I added a test to cover the added domain method.

ScottGuymer:
OK i think i have tackled all the review comments. I added to the implementation so it skips the CF header for internally routed instanced.

I think we are about ready to go with this.

eaglerainbow:
I triggered the voter; let's see what it says...

eaglerainbow (approved):

PR #196, opened 2021-04-07
Support CF-for-k8s like metrics using Kubernetes based annotations

braunsonm:
## Current Situation
Right now you have to specifically opt in an entire org/space, or individually pick apps as targets/regex match their names. With the additions of annotations and labels in V3 CAPI APIs, it makes more sense to let developers opt their applications in without modifying the promregator configuration.

## Solution Approach
This solution strictly follows what the CloudFoundry Foundation does for CF-for-k8s, adding the same sort of support to CF-for-VMs. For information on how cf-for-k8s handles this see [here](https://github.com/cloudfoundry/cf-for-k8s-metric-examples).

Basically the idea is any application can opt into metrics with the following:
```yaml
---
applications:
- name: go-app-with-metrics
  health-check-type: process
  instances: 2
  metadata:
    annotations:
      prometheus.io/scrape: "true"
      prometheus.io/port: "2112"
      prometheus.io/path: "/metrics"
```

These standard Kubernetes annotations enable scraping. 

In this PR I have implemented the following:
- Stubbed out replacements for all the V2 API calls. Moving to V3 in the future should be simplier now
- I added the V3 APIs to the CFAccessor interface instead of a V3 specific interface. This makes it easier to implement and follows along with what the CF Java Client already does (tacking `<methodname>V3` to their V3 endpoints)
- I've added a new configuration parameter to the targets listing called `kubernetesAnnotations`. Enabling this feature will still first apply all the other target matching before finally checking for the standard Kubernetes annotations on the app.
- If kubernetes annotations is enabled but V3 API support is not detected on the foundation, it is silently ignored with an warning log output
- If the developer specified the `prometheus.io/path`, Promregator will use this metrics path for this specific app. This allows every app to specify their own metrics pathing without updating the promregator configuration
- Currently `prometheus.io/port` is ignored. I have documented this.
- If kubernetes annotations is enabled and the `prometheus.io/scrape` does not equal `"true"` the application will not be scraped or output in the `/discovery` calls.
- A small tweak was made to the `HttpClient` constructed for the proxy calls to use the system client. This allows users to provide their own SSL Context on the JVM command line and gives Promregator mTLS support for free. This change is backwards compatible.

Example config:
```yaml
  targets:
    - orgName: test-org
      kubernetesAnnotations: true
```

One issue I noticed unrelated to these changes:
- During tests I had intermittent failures of `CFAccessorCacheClassicTest`. I dug into this and it seems like in `AutoRefreshingCacheMap` another thread is encountering a race condition where cache is added to `new EntryProperties` but before another call can be made by the test, the refresh thread clears it because it is not marked as "recently used" yet. This causes the test to pass sometimes and fail others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ Here is the list of major features provided by Promregator:
 * [Additional metrics are provided](docs/enrichment.md) supporting you to **monitor Promregator** and the **communication to the Cloud Foundry applications**.
 * *(>= 0.4.0)* **[Cache Invalidation](docs/invalidate-cache.md)** is possible via an (optionally auth-protected) HTTP REST endpoint.
 * Promregator's endpoints (e.g. `/metrics`, `/promregatorMetrics`, `/discovery`) support **GZIP compression**, if the clients indicates to accept it.
+* Filtering by annotations and using annotations to specify the metrics path.
 
 
 ## Architecture
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Here is the list of major features provided by Promregator:
   The available authentication schemes are easily extensible.
   
   *(>= 0.4.0)* Each target may be configured to use its own authentication scheme, thus you may authenticate to multiple CF apps using different credentials.
+  *(>= 0.9.0)* Ability to specify the [SSL Context](docs/mtls-ssl-contexts.md) used by promregator using JVM options
 * *(>= 0.2.0)* **Support for inbound authentication** (e.g. Prometheus authenticates to Promregator) using Basic HTTP Authentication. 
 * **Configuration using standard Spring properties** as defined by the Spring Framework (e.g. using `application.yml` file).
 * *(>= 0.6.0)* Support for [**encrypted passwords in configuration files**](./docs/passwords-in-config.md) including providing the encryption key via **Docker Secrets**
@@ -61,7 +62,6 @@ Here is the list of major features provided by Promregator:
 * *(>= 0.4.0)* **[Cache Invalidation](docs/invalidate-cache.md)** is possible via an (optionally auth-protected) HTTP REST endpoint.
 * Promregator's endpoints (e.g. `/metrics`, `/promregatorMetrics`, `/discovery`) support **GZIP compression**, if the clients indicates to accept it.
 * *(>= 0.9.0)* [Filtering by annotations](docs/annotation-driven.md) and using annotations to specify the metrics path.
-* *(>= 0.9.0)* Ability to specify the [SSL Context](docs/mtls-ssl-contexts.md) used by promregator using JVM options
 
 
 ## Architecture
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Here is the list of major features provided by Promregator:
 * [Additional metrics are provided](docs/enrichment.md) supporting you to **monitor Promregator** and the **communication to the Cloud Foundry applications**.
 * *(>= 0.4.0)* **[Cache Invalidation](docs/invalidate-cache.md)** is possible via an (optionally auth-protected) HTTP REST endpoint.
 * Promregator's endpoints (e.g. `/metrics`, `/promregatorMetrics`, `/discovery`) support **GZIP compression**, if the clients indicates to accept it.
-* Filtering by annotations and using annotations to specify the metrics path.
+* *(>= 0.9.0)* Filtering by annotations and using annotations to specify the metrics path.
 
 
 ## Architecture
--- a/README.md
+++ b/README.md
@@ -60,7 +60,8 @@ Here is the list of major features provided by Promregator:
 * [Additional metrics are provided](docs/enrichment.md) supporting you to **monitor Promregator** and the **communication to the Cloud Foundry applications**.
 * *(>= 0.4.0)* **[Cache Invalidation](docs/invalidate-cache.md)** is possible via an (optionally auth-protected) HTTP REST endpoint.
 * Promregator's endpoints (e.g. `/metrics`, `/promregatorMetrics`, `/discovery`) support **GZIP compression**, if the clients indicates to accept it.
-* *(>= 0.9.0)* Filtering by annotations and using annotations to specify the metrics path.
+* *(>= 0.9.0)* [Filtering by annotations](docs/annotation-driven.md) and using annotations to specify the metrics path.
+* *(>= 0.9.0)* Ability to specify the [SSL Context](docs/mtls-ssl-contexts.md) used by promregator using JVM options
 
 
 ## Architecture
--- a/docs/annotation-driven.md
+++ b/docs/annotation-driven.md
@@ -1,0 +1,33 @@
+# Annotation-Driven Configuration
+
+Starting with version 0.9.0 of Promregator, annotation-driven configuration of targets is supported if the underlying Cloud Foundry platform supports the CAPI V3 interface. Note that in these days the latter should not be a challenging requirement given the fact that the former interface, CAPI V2, has been deprecated.
+
+Configuration follows the approach suggested by [cf-for-k8s](https://github.com/cloudfoundry/cf-for-k8s-metric-examples). See also that page fro examples how to set annotations.
+
+These annotations are supported:
+
+| Annotation | supported | Comments |
+|------------|-----------|----------|
+| `prometheus.io/scrape` | yes | must be set to `"true"` |
+| `prometheus.io/path` | yes | overwrites any other `path` attribute in Promregator's main configuration, if provided |
+| `prometheus.io/port` | no | |
+
+For the metadata to be evaluated, the corresponding target needs to have set the attribute `kubernetesAnnotations` to `true` (see also details at the [configuration option page](./config.md) for this option). Note that if this option is set, any target which does *not* provide the necessary metadata annotations will not be selected for scraping (but ignored).
+
+## Example application
+
+An example Cloud Foundry application using these annotations would be deployed
+using a sample manifest like below:
+
+```yaml
+---
+applications:
+- name: go-app-with-metrics
+  metadata:
+    annotations:
+      prometheus.io/scrape: "true"
+      prometheus.io/path: "/not/default/metrics"
+```
+
+Given this configuration, the app would be scraped by promregator and the path
+for scraping would be `/not/default/metrics`.
--- a/docs/annotation-driven.md
+++ b/docs/annotation-driven.md
@@ -14,6 +14,8 @@ These annotations are supported:
 
 For the metadata to be evaluated, the corresponding target needs to have set the attribute `kubernetesAnnotations` to `true` (see also details at the [configuration option page](./config.md) for this option). Note that if this option is set, any target which does *not* provide the necessary metadata annotations will not be selected for scraping (but ignored).
 
+This provides the opportunity for an "opt-in" to scraping by applications without the need of modifying Promregator's configuration: If another application wants to get scraped, it may set `prometheus.io/scrape` to `true` (and provide a `prometheus.io/path` if applicable). On the next metadata update, the new application will be detected automatically and added to discovery.
+
 ## Example application
 
 An example Cloud Foundry application using these annotations would be deployed
--- a/docs/config.md
+++ b/docs/config.md
@@ -429,6 +429,16 @@ Defaults to `/metrics`, as this is the value which is suggested by Prometheus.
 
 Note that there may be frameworks out there, which expose their metrics in a different format using the same path `/metrics`, though. 
 
+#### Item property "promregator.targets[].kubernetesAnnotations" (optional)
+Enables support for the de facto standard Kubernetes Prometheus annotations on your CF applications. This allows each application to "opt-in" to scraping
+by specifying the annotation `prometheus.io/scrape: "true"`. Annotations support requires a version of Cloud Foundry with the V3 API otherwise this setting will be ignored. 
+
+For an example on deploying an app with these annotations see [here](https://github.com/cloudfoundry/cf-for-k8s-metric-examples).
+
+Defaults to `false`. Enabling this feature will override the `promregator.targets[].path` when it is specified using the annotation `prometheus.io/path: "/some/other/metrics"`
+
+Does not currently support the `prometheus.io/port` annotation.
+
 #### Item property "promregator.targets[].protocol" (optional)
 Specifies the protocol (`http` or `https`) which shall be used to retrieve the metrics.
 
--- a/docs/config.md
+++ b/docs/config.md
@@ -427,7 +427,8 @@ Note that the data returned by this endpoint of the target must comply to the [T
 
 Defaults to `/metrics`, as this is the value which is suggested by Prometheus.
 
-Note that there may be frameworks out there, which expose their metrics in a different format using the same path `/metrics`, though. 
+Note that there may be frameworks out there, which expose their metrics in a different format using the same path `/metrics`, though.
+Note that setting `promregator.targets[].kubernetesAnnotations` to `true` will cause that any value set to this parameter is ignored by applications that have annotation `prometheus.io/path` set.
 
 #### Item property "promregator.targets[].kubernetesAnnotations" (optional)
 Enables support for the de facto standard Kubernetes Prometheus annotations on your CF applications. This allows each application to "opt-in" to scraping
--- a/src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
+++ b/src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
@@ -17,7 +17,6 @@ import org.cloudfoundry.promregator.cfaccessor.CFAccessorSimulator;
 import org.cloudfoundry.promregator.cfaccessor.CFWatchdog;
 import org.cloudfoundry.promregator.cfaccessor.ReactiveCFAccessorImpl;
 import org.cloudfoundry.promregator.config.ConfigurationValidations;
-import org.cloudfoundry.promregator.config.validations.CompatibleCAPIVersion;
 import org.cloudfoundry.promregator.discovery.CFMultiDiscoverer;
 import org.cloudfoundry.promregator.internalmetrics.InternalMetrics;
 import org.cloudfoundry.promregator.lifecycle.InstanceLifecycleHandler;
--- a/src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
+++ b/src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
@@ -17,6 +17,7 @@ import org.cloudfoundry.promregator.cfaccessor.CFAccessorSimulator;
 import org.cloudfoundry.promregator.cfaccessor.CFWatchdog;
 import org.cloudfoundry.promregator.cfaccessor.ReactiveCFAccessorImpl;
 import org.cloudfoundry.promregator.config.ConfigurationValidations;
+import org.cloudfoundry.promregator.config.validations.CompatibleCAPIVersion;
 import org.cloudfoundry.promregator.discovery.CFMultiDiscoverer;
 import org.cloudfoundry.promregator.internalmetrics.InternalMetrics;
 import org.cloudfoundry.promregator.lifecycle.InstanceLifecycleHandler;
--- a/src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeine.java
+++ b/src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeine.java
@@ -14,6 +14,8 @@ import org.cloudfoundry.client.v2.organizations.ListOrganizationDomainsResponse;
 import org.cloudfoundry.client.v2.organizations.ListOrganizationsResponse;
 import org.cloudfoundry.client.v2.spaces.GetSpaceSummaryResponse;
 import org.cloudfoundry.client.v2.spaces.ListSpacesResponse;
+import org.cloudfoundry.client.v3.applications.ListApplicationRoutesResponse;
+import org.cloudfoundry.client.v3.spaces.GetSpaceResponse;
 import org.cloudfoundry.promregator.internalmetrics.InternalMetrics;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -263,6 +265,53 @@ public class CFAccessorCacheCaffeine implements CFAccessorCache {
 	@Override
 	public Mono<ListOrganizationDomainsResponse> retrieveAllDomains(String orgId) {
 		return Mono.fromFuture(this.domainsInOrgCache.get(orgId));
+	}
+
+	@Override
+	public Mono<org.cloudfoundry.client.v3.organizations.ListOrganizationsResponse> retrieveOrgIdV3(String orgName) {
+		// TODO: Implement cache
+		return this.parent.retrieveOrgIdV3(orgName);
+	}
+
+	@Override
+	public Mono<org.cloudfoundry.client.v3.organizations.ListOrganizationsResponse> retrieveAllOrgIdsV3() {
+		// TODO: Implement cache
+		return this.parent.retrieveAllOrgIdsV3();
+	}
+
+	@Override
+	public Mono<org.cloudfoundry.client.v3.spaces.ListSpacesResponse> retrieveSpaceIdV3(String orgId, String spaceName) {
+		// TODO: Implement cache
+		return this.parent.retrieveSpaceIdV3(orgId, spaceName);
+	}
+
+	@Override
+	public Mono<org.cloudfoundry.client.v3.spaces.ListSpacesResponse> retrieveSpaceIdsInOrgV3(String orgId) {
+		// TODO: Implement cache
+		return this.parent.retrieveSpaceIdsInOrgV3(orgId);
+	}
+
+	@Override
+	public Mono<org.cloudfoundry.client.v3.applications.ListApplicationsResponse> retrieveAllApplicationIdsInSpaceV3(String orgId, String spaceId) {
+		// TODO: Implement cache
+		return this.parent.retrieveAllApplicationIdsInSpaceV3(orgId, spaceId);
+	}
+
+	@Override
+	public Mono<GetSpaceResponse> retrieveSpaceV3(String spaceId) {
+		// TODO: Implement cache
+		return this.parent.retrieveSpaceV3(spaceId);
+	}
+
+	@Override
+	public Mono<org.cloudfoundry.client.v3.organizations.ListOrganizationDomainsResponse> retrieveAllDomainsV3(String orgId) {
+		// TODO: Implement cache
+		return this.parent.retrieveAllDomainsV3(orgId);
+	}
+
+	@Override
+	public Mono<ListApplicationRoutesResponse> retrieveRoutesForAppId(String appId) {
+		throw new UnsupportedOperationException();
 	}
 
 	@Override
--- a/src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeine.java
+++ b/src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeine.java
@@ -39,7 +39,8 @@ public class CFAccessorCacheCaffeine implements CFAccessorCache {
 	private AsyncLoadingCache<String, ListSpacesResponse> spaceIdInOrgCache;
 	private AsyncLoadingCache<CacheKeyAppsInSpace, ListApplicationsResponse> appsInSpaceCache;
 	private AsyncLoadingCache<String, GetSpaceSummaryResponse> spaceSummaryCache;		
-	private AsyncLoadingCache<String, ListOrganizationDomainsResponse> domainsInOrgCache;	
+	private AsyncLoadingCache<String, ListOrganizationDomainsResponse> domainsInOrgCache;
+	private AsyncLoadingCache<CacheKeyAppsInSpace, org.cloudfoundry.client.v3.applications.ListApplicationsResponse> appsInSpaceV3Cache;
 	
 	@Value("${cf.cache.timeout.org:3600}")
 	private int refreshCacheOrgLevelInSeconds;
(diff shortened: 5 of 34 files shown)


src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
```diff
@@ -401,6 +422,40 @@ public void setResolvedApplicationId(String resolvedApplicationId) {
 			return itnew;
 		});
 	}
+
+	private Flux<IntermediateTarget> resolveAnnotations(IntermediateTarget it) {
+		if (Boolean.TRUE.equals(it.getConfigTarget().getKubernetesAnnotations())) {
+			Mono<org.cloudfoundry.client.v3.applications.ListApplicationsResponse> response = this.cfAccessor
+				.retrieveAllApplicationsInSpaceV3(it.getResolvedOrgId(), it.getResolvedSpaceId());
```
eaglerainbow, line 429: By this way, we are doubling all space-level requests to the CFCC, if the foundation is supporting V3. Or to put it into other words, `resolveApplication` triggers a call to `retrieveAllApplicationIdsInSpace` (V2) and afterwards this logic here will  call `retrieveAllApplicationsInSpaceV3` again. However, mainly these two calls should merely provide the same response - a lot of redundant information is sent over the wire.
To that end I currently wonder whether we shouldn't merge `resolveAnnotations` into `resolveApplication`: If V3 is available, we take `retrieveAllApplicationsInSpaceV3`; if not we fall back to the old `retrieveAllApplicationIdsInSpace`. 

Disclaimer: Didn't check if that was feasible, just thinking out loud here.

docs/config.md
```diff
@@ -429,6 +429,16 @@ Defaults to `/metrics`, as this is the value which is suggested by Prometheus.
 
 Note that there may be frameworks out there, which expose their metrics in a different format using the same path `/metrics`, though.
```
eaglerainbow, line 431: Something along the lines of:
```suggestion

Note that setting `promregator.targets[].kubernetesAnnotations` to `true` will cause that any value set to this parameter is ignored by applications that have annotation `prometheus.io/path` set.
```

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
```diff
@@ -401,6 +422,40 @@ public void setResolvedApplicationId(String resolvedApplicationId) {
...
+				}
+
+				return res.getResources().stream()
+						  .filter(app -> it.getResolvedApplicationName()
+										   .equals(app.getName().toLowerCase(Locale.ENGLISH)))
+						  .filter(app -> this.isApplicationInScrapableState(app.getState().toString()))
+						  .filter(app -> app.getMetadata() != null)
+						  .filter(app -> app.getMetadata().getAnnotations() != null)
+						  .filter(app -> app.getMetadata().getAnnotations()
+											.getOrDefault("prometheus.io/scrape", "false")
```
eaglerainbow, line 445: Minor: define a constants for `"prometheus.io/scrape"` here.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
```diff
@@ -401,6 +422,40 @@ public void setResolvedApplicationId(String resolvedApplicationId) {
...
+											.equals("true"))
+						  .map(app -> {
+							  it.setResolvedMetricsPath(app.getMetadata().getAnnotations()
+														   .getOrDefault("prometheus.io/path", null));
```
eaglerainbow, line 449: Minor: define a constant for `"prometheus.io/path"` here.

eaglerainbow (commented):
Kudos for your efforts! You had a very close look on how things are done in Promregator and what my programming style is. Nice work! :)


Disclaimer: I "only" had a look at the productive coding so far - I didn't look at the unit test coding yet. 

I totally like the approach how you tackle it. The fact that we have to implement the infoV3 endpoint manually is a pitty, but maybe can be resolved at a later point in time.
You found the right spots for documentation, but I think we should add another page where we give a concrete example (call it a tiny tutorial or so). 
The fact that we don't support `prometheus.io/port` isn't critical. In the future, however, we might want to combine things with internal domains support (see also #195). 

From a coding perspective, I have the impression that this PR is doing three things at once:
1. Introduction of V3 endpoint support in the CFAccessor: There you do much more than you have to do (for the aim you really want to achieve). Though I value your work you do, this implementation is half-way only and still leaves some open ends (for example see the "// TODO: Implement cache").
2. Adding the annotiation feature, building things on item 1, which rightfully touches the `ReactiveTargetResolver` (I have left some comments there).
3. Initializing the HttpClient using `HttpClients.createSystem()`, which is a nice idea and IMO deserves a little more documentation (which kind of nice things you may do with it) to allow our users to leverage its benefits.  I little fear though that documenting this functionality also would would fix Promregator to also use HttpClients in the future (background: There were already thoughts some time back to replace the current logic with netty for introducing asynchronous HTTP communication -- but that approach hasn't manifested in anything productive yet)

Just thinking aloud: Looking at item 1, I am currently asking myself whether it wouldn't be better to split this PR into two (or even three) parts. On the item 1, we are coding "on stock" - I am unsure whether then leaving off the other V3 endpoints yet is a good or a bad idea. 

Food for the brain: The implementation shows that there might be two `ReactiveTargetResolver`s in the future: Today's one for CAPI V2, and one for CAPI V3. On bootup, after we know which version the CFCC is supporting, we could simply decide which of the two we take. Considering application annotations then would only be available in the "CAPI V3 `ReactiveTargetResolver`".

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
```diff
@@ -401,6 +422,40 @@ public void setResolvedApplicationId(String resolvedApplicationId) {
...
+
+	private Flux<IntermediateTarget> resolveAnnotations(IntermediateTarget it) {
+		if (Boolean.TRUE.equals(it.getConfigTarget().getKubernetesAnnotations())) {
+			Mono<org.cloudfoundry.client.v3.applications.ListApplicationsResponse> response = this.cfAccessor
+				.retrieveAllApplicationsInSpaceV3(it.getResolvedOrgId(), it.getResolvedSpaceId());
+
+			return response.flatMap(res -> {
+				if (res == null || INVALID_APPLICATIONS_RESPONSE == res) {
+					logEmptyTarget
+						.warn("Your foundation does not support V3 APIs, yet you have enabled Kubernetes Annotation filtering. Ignoring annotation filtering.");
```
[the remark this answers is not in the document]
braunsonm: Good point. I'll set this to debug and tweak the error message where we call the Info endpoint.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
```diff
@@ -401,6 +422,40 @@ public void setResolvedApplicationId(String resolvedApplicationId) {
...
+					return Mono.just(it);
+				}
+
+				return res.getResources().stream()
```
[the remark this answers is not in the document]
braunsonm: I wasn't sure how to reformat this. I'm not super familiar with Flux. I could `Flux.fromIterable(res.getResources())` but at the end of this logic I want to get the first element or else return an empty Mono. I can `blockFirst()` with a timeout I guess? But that didn't seem very clean either.

Open to suggestions

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFAccessorCacheCaffeine.java
```diff
@@ -37,7 +39,8 @@
 	private AsyncLoadingCache<String, ListSpacesResponse> spaceIdInOrgCache;
 	private AsyncLoadingCache<CacheKeyAppsInSpace, ListApplicationsResponse> appsInSpaceCache;
 	private AsyncLoadingCache<String, GetSpaceSummaryResponse> spaceSummaryCache;		
-	private AsyncLoadingCache<String, ListOrganizationDomainsResponse> domainsInOrgCache;	
+	private AsyncLoadingCache<String, ListOrganizationDomainsResponse> domainsInOrgCache;
+	private AsyncLoadingCache<CacheKeyAppsInSpace, org.cloudfoundry.client.v3.applications.ListApplicationsResponse> appsInSpaceV3Cache;
```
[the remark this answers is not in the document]
braunsonm: Good catch!

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFPaginatedRequestFetcher.java
```diff
@@ -40,62 +40,63 @@ public ReactiveCFPaginatedRequestFetcher(InternalMetrics internalMetrics, double
 		super();
 		this.internalMetrics = internalMetrics;
 		this.initialBackoffDelay = backoffDelay;
-		
+
 		if (requestRateLimit <= 0.0f) {
 			this.cfccRateLimiter = RateLimiter.create(Double.POSITIVE_INFINITY);
-		} else {
+		}
+		else {
```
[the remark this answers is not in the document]
braunsonm: Do you have a code style checked in somewhere? Had a hard time following the same code style. This must have been reformatted by intellij with their default settings....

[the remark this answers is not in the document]
braunsonm: Went through this and adjusted things

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
braunsonm: It's important to note we are doubling space level requests to CFCC from 1 to 2 if you have the annotation filtering enabled yes. I agree this can probably be tweaked to selectively call each endpoint depending on the CAPI version you support. I was concerned of moving core functionality to rely on the V3 APIs though and I knew this PR was already getting pretty big

braunsonm:
Hey thanks for the detailed reply!

For the tutorial, what would you like to see in it? For instance in the config docs I linked to the repository that CF has which has a demo application. That same application would work with Promregator and demonstrate the functionality

Internal domain support would be great!

You're right that this stubbed out more V3 APIs and left the caches as unimplemented. I debated doing this at all but since we will now be relying on Promregator in production I wanted to get started on the V3 support. I knew this couldn't be all done at once, especially not without you taking a look first. But I also didn't want to just leave this as a one off V3 API that is supported. Imo the idea would be to move Promregator fully onto V3 in the somewhat near future with the V2 APIs all being deprecated now. I was hoping stubbing some of this out would motivate contributors to help work on fully supporting V3

I'm glad you like the createSystem change. I saw it as a quick win for Promregator and we were wanting mTLS support. I can document something about this. There's usually 4 JVM properties you want to specify when starting promregator if you want to use this functionality. What was the alternative to `HttpClient`s that you were considering in the future? Would they not similarly support the SSL Context specified to the JVM? For instance in Spring, RestTemplates default to `createSystem()` which means this change makes us consistent with Spring's defaults.

Regarding the `ReactiveTargetResolver` separate implementations, I could see that too. As I noted in the code the space summary API differs greatly in V2 and V3. All the routes/domains have been reworked as well. I think there will be significant differences in the resolving on V3 that would make sense to provide different components for the CAPI version your foundation supports.

Are you happy with the method I used to detect V3 API support? It's unfortunate the info endpoint is not yet supported in the Java client.

braunsonm:
Added a short page about SSL Contexts. Let me know what you think.

eaglerainbow:
I saw your response. Unfortunately, although I had hoped for finding time to have another look over the weekend, it was so filled up with other stuff that I couldn't make it. As soon as time permits, however, I will get back to you...

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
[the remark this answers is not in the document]
eaglerainbow: I had a closer look at this topic and came up with ff4b24c3b7a0046cc39a10f92aa2f6bb12274c45. 
To my mind, it does not bring up a lot of value to either stream here or to add too much of reactive programming sugar. What we are doing here is more or less just working on a set of objects that is anyhow in main memory. So, I would prefer to shout out for applying the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle).

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
```diff
@@ -220,6 +229,18 @@ private void constructCloudFoundryClient() throws ConfigurationException {
...
 			log.info("Target CF platform is running on API version {}", apiVersion);
 		}
+
+		// Ensures v3 API exists. The CF Java Client does not yet implement the info endpoint for V3, so we do it manually.
+		JsonNode v3Info = new ReactorInfoV3(this.cloudFoundryClient.getConnectionContext(), this.cloudFoundryClient.getRootV3(),
+											this.cloudFoundryClient.getTokenProvider(), this.cloudFoundryClient.getRequestTags())
+			.get().onErrorReturn(JsonNodeFactory.instance.nullNode()).block();
+
+		if (v3Info == null || v3Info.isNull()) {
+			log.warn("unable to get v3 info endpoint of CF platform, some features will not work as expected");
```
eaglerainbow, line 239: minor
```suggestion
			log.warn("Unable to get v3 info endpoint of CF platform, some features will not work as expected");
```

eaglerainbow, line 239: That may be a little weak.. I see your intention, but what should the common user of Promregator do now? He will just be scared and unsure what the system does and how it behaves. 

My initial ideas at https://github.com/promregator/promregator/pull/196#discussion_r609990524 had gone into the direction of implementing a https://github.com/promregator/promregator/blob/576b5f894b883109270aff67146ab85d120b8cc2/src/main/java/org/cloudfoundry/promregator/config/validations/ConfigurationValidation.java.  As https://github.com/promregator/promregator/blob/576b5f894b883109270aff67146ab85d120b8cc2/src/main/java/org/cloudfoundry/promregator/config/ConfigurationValidations.java is a spring bean, it should be possible to also `Autowire` the `ReactiveCFAccessor` whose interface could be extended by your `v3Enabled`. I expect it won't be the last time that we will run into the situation that "some feature" won't run without v3 available.

eaglerainbow:
> For the tutorial, what would you like to see in it? [...]

After having tested it practically, I don't know whether "tutorial" really is right word for it. However, where is see the problem is at the discovery of your feature. It would be total pity, if we had such support, but no user of Promregator would find it! Only mentioning it in the release notes seems to be too weak to me. That is why I came up with 36e5c7a14bb11561bfada5ba9a6422e0c0e957f6 - but I am also not really convinced about that. At least the file can be linked from the `README.md` and thus is a little more prominent..

eaglerainbow:
> Internal domain support would be great!

Yeah, but perhaps nothing for this PR ... :)

eaglerainbow:
>  Imo the idea would be to move Promregator fully onto V3 in the somewhat near future with the V2 APIs all being deprecated now. I was hoping stubbing some of this out would motivate contributors to help work on fully supporting V3.

Hope springs eternal.  ;)
Yes, V3 is the future way to go, but I fear that these stubs may end up being loose ends. Let's keep them in for now...

eaglerainbow:
> What was the alternative to HttpClients that you were considering in the future? Would they not similarly support the SSL Context specified to the JVM?

We were thinking about using [reactor-netty](https://github.com/reactor/reactor-netty). With that, resource consumption of Promregator could even be reduced further as we wouldn't have to have all the threads synchronously waiting for scraping. I haven't had a deep look at the SSL support, but given that netty is much more low-level in its thinking than all the frameworks discussed here, I fear that it will be "significantly different".

eaglerainbow:
> Are you happy with the method I used to detect V3 API support? It's unfortunate the info endpoint is not yet supported in the Java client.

Indeed. But from my perspective it's good enough for our purpose - and that's what counts.
I also saw that you had raised https://github.com/cloudfoundry/cf-java-client/issues/1097 for that. Let's hope that there is some movement.

eaglerainbow:
> Added a short page about SSL Contexts. Let me know what you think.

Interesting. Some formatting / editing might be good, but that's just doing it. 
As these are native Java properties, I currently fear that our docker consumers may have troubles finding out that setting the `JAVA_OPTS` environment variable to their container permits them to configure the SSL context without much of a-do. Perhaps that also worth to mention in your documentation file?
Also, we should mention your mTLS support through default SSL Context in the feature list of README.md. Otherwise we might run again into trouble that your nice feature is overlooked.

WDYT?

checketts:
I've been loosely following this thread. Does CF need to be running on k8s to leverage this feature? If not, I'm interested in trying it out.

braunsonm:
@checketts Nope, this is meant to work with regular CF-for-VMs deployments like bosh. But this way if you ever do move to cf-for-k8s, the behaviour is consistent. (And imo better as it allows devs to opt in). Please do give it a shot! If you're interested in a beta test you can grab a jar from my forked repo in the releases.

@eaglerainbow will get back to you on your comments tomorrow!

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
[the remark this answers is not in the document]
braunsonm: In my opinion a stream is going to be simpler than doing a bunch of loops and if statements.

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
braunsonm: I've implemented a validator. I did some refactoring to the existing ones to just be beans so it's easier to extend these validators to use other beans in the future.

braunsonm:
> After having tested it practically, I don't know whether "tutorial" really is right word for it. However, where is see the problem is at the discovery of your feature. It would be total pity, if we had such support, but no user of Promregator would find it! Only mentioning it in the release notes seems to be too weak to me. That is why I came up with 36e5c7a - but I am also not really convinced about that. At least the file can be linked from the README.md and thus is a little more prominent..

I like this. I included it with an example manifest for a CF application.

> As these are native Java properties, I currently fear that our docker consumers may have troubles finding out that setting the JAVA_OPTS environment variable to their container permits them to configure the SSL context without much of a-do. Perhaps that also worth to mention in your documentation file?
Also, we should mention your mTLS support through default SSL Context in the feature list of README.md. Otherwise we might run again into trouble that your nice feature is overlooked.

Added!

Regarding netty that's a good thing to keep in mind. If you have an ongoing issue for adding support it would be smart to add a note about investigating mTLS. It would be necessary to support specifying a truststore/keystore.

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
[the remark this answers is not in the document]
braunsonm: Let me know if you'd rather have streams or I saw you did a commit to a branch I could bring in instead.

eaglerainbow:
> Regarding netty that's a good thing to keep in mind. If you have an ongoing issue for adding support it would be smart to add a note about investigating mTLS. It would be necessary to support specifying a truststore/keystore.

So far this is only some local test branch I have on my local machine. In this prototype it turned out that things would be a lot more complicated and would include incompatible changes to be made so that I refrained so far from doing it. As the main benefit of it would be performance/TCO and nobody has experienced issues for that part of Promregator so far (at least not publicly reported), it never found its path to light.

README.md
```diff
@@ -60,6 +60,8 @@ Here is the list of major features provided by Promregator:
 * [Additional metrics are provided](docs/enrichment.md) supporting you to **monitor Promregator** and the **communication to the Cloud Foundry applications**.
 * *(>= 0.4.0)* **[Cache Invalidation](docs/invalidate-cache.md)** is possible via an (optionally auth-protected) HTTP REST endpoint.
 * Promregator's endpoints (e.g. `/metrics`, `/promregatorMetrics`, `/discovery`) support **GZIP compression**, if the clients indicates to accept it.
+* *(>= 0.9.0)* [Filtering by annotations](docs/annotation-driven.md) and using annotations to specify the metrics path.
+* *(>= 0.9.0)* Ability to specify the [SSL Context](docs/mtls-ssl-contexts.md) used by promregator using JVM options
```
eaglerainbow, line 64: I think that one should better to go to line 51 and become a subitem to "Support of Authentication Schemes" (just from logic's point of view)

docs/annotation-driven.md
```diff
@@ -0,0 +1,33 @@
...
+These annotations are supported:
+
+| Annotation | supported | Comments |
+|------------|-----------|----------|
+| `prometheus.io/scrape` | yes | must be set to `"true"` |
+| `prometheus.io/path` | yes | overwrites any other `path` attribute in Promregator's main configuration, if provided |
+| `prometheus.io/port` | no | |
+
+For the metadata to be evaluated, the corresponding target needs to have set the attribute `kubernetesAnnotations` to `true` (see also details at the [configuration option page](./config.md) for this option). Note that if this option is set, any target which does *not* provide the necessary metadata annotations will not be selected for scraping (but ignored).
+
```
eaglerainbow, line 16: Just came to my mind:

```suggestion
This provides the opportunity for an "opt-in" to scraping by applications without the need of modifying Promregator's configuration: If another application wants to get scraped, it may set `prometheus.io/scrape` to `true` (and provide a `prometheus.io/path` if applicable). On the next metadata update, the new application will be detected automatically and added to discovery.

```

src/main/java/org/cloudfoundry/promregator/PromregatorApplication.java
```diff
@@ -17,6 +17,7 @@
 import org.cloudfoundry.promregator.cfaccessor.CFWatchdog;
 import org.cloudfoundry.promregator.cfaccessor.ReactiveCFAccessorImpl;
 import org.cloudfoundry.promregator.config.ConfigurationValidations;
+import org.cloudfoundry.promregator.config.validations.CompatibleCAPIVersion;
```
eaglerainbow, line 20: necessary?

src/main/java/org/cloudfoundry/promregator/scanner/ReactiveTargetResolver.java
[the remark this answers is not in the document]
eaglerainbow: For me `if` and `for` are simpler than streams, but I am not religious about that. We may keep it as it is. We don't lose/win much either way (I had thought it would have been more before).

eaglerainbow (commented):
Found some more mainly minor stuff

eaglerainbow:
Voter currently runs red due to 

``` 
CFAccessorSimulatorTest.testRetrieveOrgIdV3()

org.opentest4j.AssertionFailedError: Expected java.lang.UnsupportedOperationException to be thrown, but nothing was thrown.
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:71)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:37)
	at org.junit.jupiter.api.Assertions.assertThrows(Assertions.java:3007)
	at org.cloudfoundry.promregator.cfaccessor.CFAccessorSimulatorTest.testRetrieveOrgIdV3(CFAccessorSimulatorTest.java:96)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:64)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:688)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:149)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:140)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:84)
	at org.junit.jupiter.engine.execution.ExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(ExecutableInvoker.java:115)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.lambda$invoke$0(ExecutableInvoker.java:105)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.proceed(InvocationInterceptorChain.java:64)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.chainAndInvoke(InvocationInterceptorChain.java:45)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.invoke(InvocationInterceptorChain.java:37)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:104)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:98)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$6(TestMethodTestDescriptor.java:210)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeTestMethod(TestMethodTestDescriptor.java:206)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:131)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:65)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:129)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:127)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:126)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:84)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:38)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:143)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:129)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:127)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:126)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:84)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:38)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$5(NodeTestTask.java:143)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$7(NodeTestTask.java:129)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:127)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:126)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:84)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.submit(SameThreadHierarchicalTestExecutorService.java:32)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor.execute(HierarchicalTestExecutor.java:57)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestEngine.execute(HierarchicalTestEngine.java:51)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:229)
	at org.junit.platform.launcher.core.DefaultLauncher.lambda$execute$6(DefaultLauncher.java:197)
	at org.junit.platform.launcher.core.DefaultLauncher.withInterceptedStreams(DefaultLauncher.java:211)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:191)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:137)
	at org.eclipse.jdt.internal.junit5.runner.JUnit5TestReference.run(JUnit5TestReference.java:89)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:41)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:542)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:770)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:464)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:210)
```

Will have a look into this topic later that week...

braunsonm:
Not sure how I missed that one thanks. All above fixed.

eaglerainbow (approved):

eaglerainbow:
Here we go!

eaglerainbow:
@braunsonm Thanks for your contribution!

braunsonm:
Hey @eaglerainbow just wondering when you're planning to cut a release? For now we'll run a snapshot from my fork in prod :)

eaglerainbow:
Seems like a good candidate then ;)

eaglerainbow:
I am currently on it..

eaglerainbow:
Here we go: https://github.com/promregator/promregator/releases/tag/v0.9.0

PR #202, opened 2021-06-17
Enable handling for custom credential vaulting

cyberseventies:
## Current Situation
Promregator depends on static cf_api credentials passed during start of application.

## Problem Statement
This behavior prevents custom vaulting of credentials (i.e. with runtime password rotations, etc., etc.).

## Solution Approach
Use a simple Spring configuration-based approach to inject an object representing the mechanism for getting the credentials.  This object can be static (the default) or can be overridden by a custom implementation.

## Prerequisites
Promregator should be published both as runnable jar and a library (pom change, classifier).

## Post Activities
Release this new version to make both the executable and the library available in Maven central.


pom.xml
```diff
@@ -279,6 +279,9 @@
 			<plugin>
 				<groupId>org.springframework.boot</groupId>
 				<artifactId>spring-boot-maven-plugin</artifactId>
+		                <configuration>
+                                        <classifier>exec</classifier>
```
eaglerainbow, line 283: Just briefly checked on this: This would be incompatible for maven central downloaders 😢 
So far, our "full-blown package" has no classifier.
With this way, the "full-blown package" would get a classifier "exec". The "promregator library" would not have the classifier. So people so far just downloading the thing without any suffix will now bump into the library instead of the "full version". We need to turn this around somehow.

I think we need to further play around with the [options of `repackage`](https://docs.spring.io/spring-boot/docs/1.1.x/reference/html/howto-build.html) here.

pom.xml
```diff
@@ -279,6 +279,9 @@
 			<plugin>
 				<groupId>org.springframework.boot</groupId>
 				<artifactId>spring-boot-maven-plugin</artifactId>
+		                <configuration>
```
eaglerainbow, line 282: Additionally, we need to get more clarity on the discussion of #197. 
It also means that we need to solve the [matter of "public solution model"](https://github.com/promregator/promregator/discussions/197#discussioncomment-743876). Perhaps adding some annotations to indicate whether some API is public or not would also be sufficient?

eaglerainbow (commented):
One more thing necessary in general: We need to adjust documentation explaining that cf.username and cf.password *may* be optional in case a different authentication method for CFCC was implemented by a "plugin"/enhancement.

src/main/java/org/cloudfoundry/promregator/cfaccessor/CFWatchdog.java
```diff
@@ -36,7 +36,7 @@
 	
 	@Scheduled(fixedRateString = "${cf.watchdog.rate:60}000", initialDelayString = "${cf.watchdog.initialDelay:60}000")
 	@SuppressWarnings("unused")
-	private void connectionWatchdog() {
+	protected void connectionWatchdog() {
```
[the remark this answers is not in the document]
cyberseventies: Upon review of our own code, I discovered we were doing something unexpected that caused visibility issues here at startup. I have since fixed our side and pushed the change back to _private_ visibility.

pom.xml
cyberseventies: This is correct.  Just plainly implementing this in this way would cause issues for those depending on getting the executable when providing no classifier at all. The supposed pseudo-standard here is to have the client request the "exec" classifier, but I can absolutely see how that would be very painful given that many are already using it.

Would you be comfortable with a classifier of ```lib``` or ```library``` (or something along those lines) for pulling the dependency jar?  This would keep behavior the same for existing clients (no classifier to get executable, full-blown jar) and only those who wish to use it as a dependency would provide the above classifier...

I am open to whatever alternative is best in this case, as long as we can also reference Promregator as a dependency.

pom.xml
cyberseventies: This is another great point in terms of maintenance and long-term flexibility. Right now, it is not immediately clear what is available to be extended / overridden safely. Clearly, this is part of a larger discussion as we need to determine what we think is going to remain a constant, stable interface and what is actually implementation details subject to change with new versions of CloudFoundry, etc., etc.

My personal experience is that we'd want to make very clear "why" (what reasons) a user would generally want to extend Promregator (i.e. custom credentials implementation, custom verification and caching of those creds, etc.).  Things outside of that list would be "at your own risk" in terms of extension - unless of course, the user opened a PR for review here.

Reference to your mention here: https://github.com/promregator/promregator/discussions/197#discussioncomment-743889, we'd need to consider the same issue for any additional library introduced.

We can absolutely use annotations as suggested or whatever solution makes the most sense. Another (higher effort) option would be package structures which make it clear what is going to remain backward compatible and what is internal.

pom.xml
```diff
@@ -276,6 +276,19 @@
 
 	<build>
 		<plugins>
+			<plugin>
```
cyberseventies, line 279: NB: this needs to be before the spring-boot-maven-plugin as both execute in the same phase.
See this [link](https://docs.spring.io/spring-boot/docs/current/maven-plugin/reference/htmlsingle/#packaging.examples.custom-classifier) for details.

eaglerainbow: I like the approach. I have added a comment on this (as this could be a source of ever-lasting questions otherwise ;) ) and pushed it to your branch.

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
```diff
@@ -150,7 +146,7 @@ private DefaultConnectionContext connectionContext(ProxyConfiguration proxyConfi
 	}
 
 	private PasswordGrantTokenProvider tokenProvider() {
-		return PasswordGrantTokenProvider.builder().password(this.password).username(this.username).build();
+		return PasswordGrantTokenProvider.builder().password(this.cfApiCredentials.getPassword()).username(this.cfApiCredentials.getUsername()).build();
```
eaglerainbow, line 149: Looking at this now, it comes to my mind that you are using a feature which we haven't built intentionally, though :) 
BTW: How you are detecting that the password expired? Just by the fact that you can't connect to CF API anymore? Would there be alternative options, e.g. event mechanism?

eaglerainbow:
I've taken the liberty to add to your branch and introduce the public solution mechanism through annotations. From a writing it feels rather good. Yet, I'd really like to hear some comments from you how you feel about it. @checketts can you also have a look here? This sounds like a rather fundamental change to me which should be thought through rather well...

docs/extension/README.md
```diff
@@ -0,0 +1,26 @@
+# Extending Promgregator
+
+Promregator supports extending it. Starting with 0.10.0, Promregator not just provides runnable artifacts like the `promregator.jar` file or the Docker image, but also `promregator-library` (available through Maven repositories) is delivered. You may include this library as a dependency to your own project, for example by using the following dependency declaration in your `pom.xml`:
+
+```xml
+<dependency>
```
cyberseventies, line 6: @eaglerainbow - The pom is setup to provide a classifier.  This configuration probably won't work. With classifiers I have usually seen the dependency defined as such:
```
<dependency>
        <groupId>com.github.promregator</groupId>
	<artifactId>promregator</artifactId>
        <classifier>library</classifier>
	<version>0.10.0</version> <!-- or any other higher version number -->
</dependency>
```
The build will produce dependencies:

- promregator-0.10.0-SNAPSHOT.jar (The executable boot jar)
- promregator-0.10.0-SNAPSHOT-library.jar (The library jar)
- promregator-0.10.0-SNAPSHOT-sources.jar (The source code jar)
and so on.

I think your configuration noted in the PR above would expect to find "promregator-library-0.10.0-SNAPSHOT.jar" in the m2 repo. Just for clarity, my local directory would be something like:
`/development/.m2/repository/com/github/promregator/promregator-library/0.10.0-SNAPSHOT/promregator-library-0.10.0-SNAPSHOT.jar`

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
cyberseventies: Unintentional features are sometime the _very best_ kind. The thought here was to leave it to the implementer of extension based on the operation of their credentials mechanism:

- always pull - get new creds every time before calling cf api
- test-before-use - ensure that cached creds are valid and pull new ones if necessary before calling cf api
- pull on failure - assume cached creds are valid and pull new ones if a 401/403 (or whatever custom response) is received from cf api
- message on change - receive notification from platform/service which causes cached credentials to be updated before calling cf api
- any number of other custom solutions which cause cached creds to be updated

docs/extension/README.md
eaglerainbow: You are right. I have pushed 78a67e3d78b8cc3b5cad0b9693d90655de5e53d9 which should make use of the classifier approach properly. Please check from your side!

src/main/java/org/cloudfoundry/promregator/cfaccessor/ReactiveCFAccessorImpl.java
eaglerainbow: Well, yes, but this decision is only possible, if the library is asking the TokenProvider in first place. I would assume that `cf-java-client` is caching the JWT, so there are only very occasional inquiries coming to the token provider (hopefully!).
I do not want to intervene in the way the implementation is doing things here - they should have their full freedom how to handle caching. It's more the fear that due to several authentication attempts, `cf-java-client` might be asking the extension *too late*.

eaglerainbow: A lot of time has passed after my last comment. The reply to my doubts are is simple: As long as the JWT is valid (i.e. "not expired"), it does not matter that the password might have changed: The JWT is still valid and no username/password is necessary anymore. 
The only thing which is important is that upon each request of a new JWT, the `cf-java-client` library needs to query the registered `PasswordGrantTokenProvider` - which is what it does.

PR #203, opened 2021-09-16
bugfix: NPE when there is an application with no routes

dkhacharia:
## Current Situation
If there is a running app which does not have any routes bound to it, we get a NPE.

## Problem Statement
In this case, the whole monitoring is stopped as it is not recoverable and promregator does not do anything.

## Solution Approach
If there are no urls, ignore the application

## Symptom
``` 
java.lang.NullPointerException: Cannot invoke "String.isEmpty()" because the return value of "org.cloudfoundry.promregator.scanner.ReactiveAppInstanceScanner$OSAVector.getDomainId()" is null
	at org.cloudfoundry.promregator.scanner.ReactiveAppInstanceScanner.lambda$10(ReactiveAppInstanceScanner.java:284) ~[classes/:na]
	at reactor.core.publisher.FluxFlatMap$FlatMapMain.onNext(FluxFlatMap.java:386) ~[reactor-core-3.4.5.jar:3.4.5]
	at reactor.core.publisher.FluxZip$ZipCoordinator.drain(FluxZip.java:756) ~[reactor-core-3.4.5.jar:3.4.5]
	at reactor.core.publisher.FluxZip$ZipInner.onNext(FluxZip.java:915) ~[reactor-core-3.4.5.jar:3.4.5]
	at reactor.core.publisher.FluxMergeSequential$MergeSequentialMain.drain(FluxMergeSequential.java:432) ~[reactor-core-3.4.5.jar:3.4.5]
```

### DIFF
--- a/src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
+++ b/src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
@@ -251,7 +251,9 @@ public class ReactiveAppInstanceScanner implements AppInstanceScanner {
 				// Set the access url to the selected route (without any protocol or path yet)
 				v.setAccessURL(this.determineApplicationRoute(urls,
 						v.getTarget().getOriginalTarget().getPreferredRouteRegexPatterns()));
-			}
+			} else
+				// if there is no url, skip this one
+				return Mono.empty();
 
 			// In the interest of backwards compatibility lest do this inside a try.
 			// There is little reason why this should not find the correct domain
@@ -274,7 +276,7 @@ public class ReactiveAppInstanceScanner implements AppInstanceScanner {
 			OSAVector v = tuple.getT1();
 			List<DomainResource> domains = tuple.getT2();
 
-			if (domains.size() == 0) {
+			if (domains.size() == 0 || v.getDomainId() == null) {
 				// NB: This drops the current target!
 				return Mono.empty();
 			}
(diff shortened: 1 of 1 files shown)


src/main/java/org/cloudfoundry/promregator/scanner/ReactiveAppInstanceScanner.java
```diff
@@ -251,7 +251,9 @@ public void setNumberOfInstances(int numberOfInstances) {
 				// Set the access url to the selected route (without any protocol or path yet)
 				v.setAccessURL(this.determineApplicationRoute(urls,
 						v.getTarget().getOriginalTarget().getPreferredRouteRegexPatterns()));
-			}
+			} else
+				// if there is no url, skip this one
+				return Mono.empty();
```
eaglerainbow, line 256:
```suggestion
			} else {
				// if there is no url, skip this one
				return Mono.empty();
			}
```

eaglerainbow (commented):
Nice catch.
I just wonder that we don't have a unit test for this (looking into it)

eaglerainbow:
most likely caused by #195

eaglerainbow:
Requires "downport" to 0.9.x plus release of new patch.

eaglerainbow:
Don't forget to merge https://github.com/promregator/promregator/tree/pr203-unittest as well (provides unit tests for this case here)

eaglerainbow:
Hi @dkhacharia 

Thanks for submitting! If you accept the small adjustment of the review, I'd merge this PR (plus some a unit test) for it and downport things to 0.9.x.

dkhacharia:
Thanks for quick response.
Suggestion accepted. Please merge it :)

eaglerainbow:
Thanks for your PR!

eaglerainbow:
89b9af085e1c11610bed1fee509341ff9b34b871 merged into `master`
and double maintained via cherry-pick into `rel-0.9`.

PR #204, opened 2021-09-17
Spring-boot & cf-client-java version bumping

eaglerainbow:
## Current Situation
Spring-boot and cf-client-java library received new releases.

## Problem Statement
We want to make use of their bugfixes.

## Solution Approach
Bump them


eaglerainbow:
Also double maintained to 0.9.1 and later

Issue #207, opened 2021-10-30
Discussion: using coroutines to simplify reactive calls


checketts:
The reactive client is powerful, but complex. 

Kotlin coroutines allow hooking into the reactive code, and continue the benefits of the async nature, while making the code simpler to read and reason about.

I've begun a first pass and it already simplfied the appscaner to no longer needs to do extra look ups (for the spaceID since we already had looked it up)

I'll be doing a bit more to see if the pagination logic can be simplified too. 

Reactive (and Java)
```java
		Mono<ListOrganizationsResponse> responseMono = this.cfAccessor.retrieveAllOrgIds();

		Flux<OrganizationResource> orgResFlux = responseMono.map(ListOrganizationsResponse::getResources)
			.flatMapMany(Flux::fromIterable);
		
		if (it.getConfigTarget().getOrgRegex() != null) {
			// Case 2
			final Pattern filterPattern = Pattern.compile(it.getConfigTarget().getOrgRegex(), Pattern.CASE_INSENSITIVE);
			
			orgResFlux = orgResFlux.filter(orgRes -> {
				Matcher m = filterPattern.matcher(orgRes.getEntity().getName());
				return m.matches();
			});
		}
		
		return orgResFlux.map(orgRes -> {
			IntermediateTarget itnew = new IntermediateTarget(it);
			itnew.setResolvedOrgId(orgRes.getMetadata().getId());
			itnew.setResolvedOrgName(orgRes.getEntity().getName());
			
			return itnew;
		});
```

```kotlin
                cfAccessor.retrieveAllOrgIds(api).awaitSingle().resources.map {
                    OrgResponse(api, it.metadata.id, it.entity.name)
                }

      //snip intermediary code

         val orgMatches = apps.filter { app -> configTargets.any{ target ->
            (target.orgName == null && target.orgRegex == null) ||
            (target.orgName == null && target.orgRegex?.toRegex()?.matches(app.space.org.name) == true) ||
                    (target.orgRegex == null && target.orgName == app.space.org.name)
        } }
```

Issue #158, opened 2020-02-21
Deprecate /metrics endpoint in favor of Consul discovery


checketts:
## Summary / Problem Statement
The `/metrics` endpoint is a convenient starting point to immediately get metrics from Promregator, since you merely point Prometheus to the Promeregator instance. However it doesn't scale and adds a lot of complexity to Promregator.

### Can't scale
* As more instances are added, the `/metrics` endpoint becomes slower since all instances have to be scraped at the same time. By default Prometheus will timeout the request after 10 seconds. A single slow instance will make all metrics disappear in Prometheus.

### Added complexity
* By scraping several instances and exposing them via a single `/metrics` endpoint Promregator has to add labels to differentiate among the instances called
  * To add labels, Promregator has to parse the existing labels
  * Some labels/metrics can't be created in this way (`up` is one example)
* Promregator has to avoid calling itself to avoid infinite loop.
  * This means Promregator can't be exposed as a typical target, and it has to be treated differently via discovery.
  * By exposing Promregator via a different mechanism, the org/space tags aren't available (see #117)

## Proposed replacement - Consul

The current recommendation is to stop using `/metrics` and use the `/discovery` endpoint. This makes adoption much harder since Prometheus can't scrape the `/discovery` endpoint directly.

A better solution that would minimize Promregator configuration and centralize target configuration/relabeling in Prometheus would be to have Prometheus call Promregator directly for discovery.

This can be done using one of the supported service discovery mechanisms. After investigating several options, the [Consul based one](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#consul_sd_config) is the simplest to implement.

Here is an example of the configuration in Prometheus:
```
scrape_configs:
  - job_name: "promregator"
    scheme: https
    consul_sd_configs:
      - server: https://mypromregator.mydomain.org  #1 
       # services:   #2
       # - my-app
    relabel_configs:
     - source_labels: [__meta_consul_service_metadata_metrics_path] #3 
       action: replace
       target_label: __metrics_path__       
     - source_labels: [__meta_consul_service,__meta_consul_service_metadata_cf_instanceNumber]
       target_label: instance
       separator: "-"
     - source_labels: [__meta_consul_service_metadata_orgName]
       target_label: org_name
     - source_labels: [__meta_consul_service_metadata_spaceName]
       target_label: space_name
     - source_labels: [__meta_consul_service]
       target_label: application
```

[...]

PR #208, opened 2021-11-03
Promregator Lite Proposal

checketts:
I apologize for the massive PR. I've been running a highly customized Promregator build for the last year, this is an attempt to begin to bring those lessons learned upstream.

This is an attempt to lighten and deprecate unneeded Promregator features.

# Feature to deprecate:

- [ ] Metrics (#158) endpoint (single target metrics should be preferred)
- [ ] Metrics enrichment (metrics should be provided for individual targets as a passthrough)
- [ ] promregator_up metric (By using passthrough the built in Prometheus `up` metric can be used
- [ ] OAuth2XSUAAEnricher isn't completely migrated (held off while focusing on the main PR)

# New features introduced:

- [X] A `lite` package has been created with duplicated code from the discovery/scanner/accessor classes. THe purpose to be able to rewrite those without impacting the current endpoints, but allowing those to be deprecated.
- [X] `/v2/discovery` endpoint with the goal of making this the primary endpoint eventually. Currently is behaves very similarly to the existing discovery, but it will evolve to leverage [Prometheus HTTP discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#http_sd_config)
- [X] `/v2/singleTargetMetrics` the doesn't enrich metrics, but instead passes through the call (optionally adding authentication config). This lowers Promregator complexity, and also allows errors in underlying targets to surface to Prometheus itself. (addresses #210 )
- [X] This PR introduces Kotlin to the codebase. Unfortunately it couldn't be avoided to due coroutine needs
- [X] Leverage coroutines (#207 ) to simplify the Reactive CF client interactions
- [X] #209 Adds support for multiple APIs


src/main/java/org/cloudfoundry/promregator/lite/discovery/RateLimiter.kt
```diff
@@ -0,0 +1,79 @@
...
+
+import io.micrometer.core.instrument.MeterRegistry
+import kotlinx.coroutines.*
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.atomic.AtomicLong
+
+private val log = mu.KotlinLogging.logger {  }
+
+class RateLimiter(
```
checketts, line 11: Since the only Guava class we use is for RateLimiting I implemented one specific to our needs. This uses the 'leaky bucket' type algorithm. 

Essentially the bucket has X tokens a minute, and it refills 1/60th of the tokens every second when it isn't full. This allows for initial bursts that might drain the bucket, then a trickle of requests as the bucket slowly refills.

It also provides 3 metrics: availableTokens, queueDepth, and totalWaitTime. Those are exposed elsewhere as metrics.

src/main/java/org/cloudfoundry/promregator/lite/discovery/SimpleAppInstanceScanner.kt
```diff
@@ -0,0 +1,109 @@
+package org.cloudfoundry.promregator.lite.discovery
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+import java.util.regex.Pattern
+
+private val log = mu.KotlinLogging.logger {  }
+
+@Component
+class SimpleAppInstanceScanner(
```
checketts, line 10: With the coroutines refactor the instance scanner no longer needs to talk to the CFAccessor, this is now a simple class that calculates the instance access URLs. I kept that logic largely the same for now.

src/main/java/org/cloudfoundry/promregator/lite/discovery/SimpleTargetResolver.kt
```diff
@@ -0,0 +1,191 @@
...
+import org.cloudfoundry.promregator.lite.config.CloudFoundryConfig
+import org.cloudfoundry.promregator.lite.config.PromregatorConfiguration.Companion.DEFAULT_ID
+import org.springframework.stereotype.Component
+import java.util.*
+import java.util.concurrent.atomic.AtomicBoolean
+import org.cloudfoundry.promregator.lite.config.Target
+import kotlin.text.RegexOption.IGNORE_CASE
+
+
+data class OrgResponse(
```
checketts, line 17: Note these data classes replace all the previous intermediate targets and other temp classes used when using the reactive approach. They are all immutable (val) properties so we can build up the results from the CF queries easily.

src/main/java/org/cloudfoundry/promregator/lite/discovery/SimpleTargetResolver.kt
```diff
@@ -0,0 +1,191 @@
...
+}
+
+private val log = mu.KotlinLogging.logger {  }
+
+@Component
+class SimpleTargetResolver(
+        private val cfAccessor: SimpleCFAccessorImpl,
+        private val cf: CloudFoundryConfig
+) {
+    suspend fun resolveTargets(configTargetsToMatch: List<Target>): List<AppResponse> = coroutineScope {
```
checketts, line 70: This resolveTargets is the main method. With the reactive code removed it is much simpler to reason about.

src/main/java/org/cloudfoundry/promregator/lite/discovery/SimpleTargetResolver.kt
```diff
@@ -0,0 +1,191 @@
...
+        val configTargets = configTargetsToMatch.ifEmpty { listOf(Target()) }
+
+        val orgs = resolveOrgs(configTargets)
+        val domains = orgs.resolveDomains()
+        val spaces = orgs.resolveSpaces(configTargets, domains)
+
+        val apps = if (cfAccessor.isV3Enabled) {
```
checketts, line 77: I look forward to expanding on the v3 usage now that it is so much easier to reason about. Now the logic to only gather the space summaries when v3 is enabled is much simpler to follow.

src/main/java/org/cloudfoundry/promregator/lite/discovery/SimpleTargetResolver.kt
```diff
@@ -0,0 +1,191 @@
...
+            val spaceSummaries = spaces.resolveSummaries()
+            spaces.resolveAppsV3(spaceSummaries)
+        } else {
+            spaces.resolveAppsV2()
+        }
+
+        return@coroutineScope apps.map { app ->
```
checketts, line 84: This method finishes with the filtering of apps based on name/regex

src/main/java/org/cloudfoundry/promregator/lite/discovery/SimpleTargetResolver.kt
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    private suspend fun List<OrgResponse>.resolveSpaces(configTargets: List<Target>, domains: List<DomainResponse>) = coroutineScope {
+        return@coroutineScope this@resolveSpaces.map { org ->
+            async {
+                cfAccessor.retrieveSpaceIdsInOrg(org.api, org.id).map { space ->
+                    SpaceResponse(org, space.entity.name, space.metadata.id, domains.filter { it.orgId == org.id })
+                }
+            }
+        }.awaitAll().flatten().filter { space ->
```
checketts, line 129: The `async` and `awaitAll` calls allow for the parallel processing that keeps this fast and easy to reason about. This is Kotlin coroutines being used. You'll notice the `suspend` keyword on the function. Coroutines yield the thread (suspending) during things like `await`ing so that is how the parallelism is managed.

src/main/java/org/cloudfoundry/promregator/lite/discovery/SimpleTargetResolver.kt
```diff
@@ -0,0 +1,191 @@
...
+                    AppResponse(appRes.id, appRes.name, space, scrapable, annotationMetricsPath,
+                            spaceAppSummary = spaceSummaries.firstOrNull { it.appName == appNameLower }?.appSummary)
+                }
+
+            }
+        }.awaitAll().flatten()
+    }
+
+    private suspend fun List<SpaceResponse>.resolveAppsV2() = coroutineScope {
+        return@coroutineScope this@resolveAppsV2.map { space ->
```
checketts, line 172: These `return@coroutineScope` syntax aren't required. I included them to help explain what is being returned for those new to Kotlin. We can remove them

src/main/java/org/cloudfoundry/promregator/lite/scrape/ScrapeControllerV2.kt
```diff
@@ -0,0 +1,65 @@
...
+@RestController
+class ScrapeControllerV2(
+        private val discoverer: DiscoveryService,
+        private val webClientBuilder: WebClient.Builder,
+        private val authService: AuthenticatorService,
+) {
+    private val httpClient = HttpClient.create()
+            .option(ChannelOption.CONNECT_TIMEOUT_MILLIS, 3000) // Connection Timeout
+            .doOnConnected { connection ->
+                connection.addHandlerLast(ReadTimeoutHandler(20)) // Read Timeout
```
checketts, line 31: This is the new scrape endpoint. It doesn't do any metric enrichment. We can make the timeouts configurable if need be.

src/main/java/org/cloudfoundry/promregator/lite/scrape/ScrapeControllerV2.kt
```diff
@@ -0,0 +1,65 @@
...
+                        .addHandlerLast(WriteTimeoutHandler(20)) // Write Timeout
+            }
+    private val webClient: WebClient = webClientBuilder
+                        .clientConnector(ReactorClientHttpConnector(httpClient))
+                        .build()
+
+    @GetMapping(value = ["/v2/singleTargetMetrics/{appId}/{instanceNumber}"])
+    fun scrape(@PathVariable appId: String,
+                       @PathVariable instanceNumber: String): ResponseEntity<String> = runBlocking {
+        val instance = discoverer.discoverCached()
```
checketts, line 41: We reuse the cache to look up the instance details. (And it will trigger a cache population if needed)

src/main/java/org/cloudfoundry/promregator/lite/scrape/ScrapeControllerV2.kt
```diff
@@ -0,0 +1,65 @@
...
+        val authEnricher = authService.getAuthenticationEnricherById(instance.target.originalTarget?.authenticatorId)
+        log.debug { "Scraping $instance" }
+
+        val respHeaders = HttpHeaders().apply {
+            set(HttpHeaders.CONTENT_TYPE, TextFormat.CONTENT_TYPE_004)
+        }
+        webClient
+                .get()
+                .uri(instance.accessUrl)
+                .header("X-CF-APP-INSTANCE", "${instance.applicationId}:${instance.instanceNumber}")
```
checketts, line 54: cf app instance header applied

src/main/java/org/cloudfoundry/promregator/lite/scrape/ScrapeControllerV2.kt
```diff
@@ -0,0 +1,65 @@
...
+                .headers { authEnricher.enrichWithAuthentication(it) }
+                .awaitExchange { resp ->
+                    if(resp.statusCode() != HttpStatus.OK) {
+                        log.error { "Error scraping ${instance.accessUrl} status:${resp.statusCode()}" }
```
checketts, line 58: If the scrape fails (bad path, auth, service down) we want to surface the status code to Prometheus itsself so it can have proper `up` metrics for each target. This eliminates the `promregator_up` metric

src/main/java/org/cloudfoundry/promregator/lite/websecurity/SecurityConfig.java
```diff
@@ -0,0 +1,163 @@
...
+import org.springframework.security.core.userdetails.UserDetailsService;
+import org.springframework.security.provisioning.InMemoryUserDetailsManager;
+import org.springframework.security.web.savedrequest.HttpSessionRequestCache;
+import org.springframework.security.web.savedrequest.RequestCache;
+
+import java.util.UUID;
+
+//@Configuration
+//@EnableWebSecurity
+public class SecurityConfig extends WebSecurityConfigurerAdapter {
```
checketts, line 24: This can be deleted and was a leftover from when the `lite` codebase was a separate project

src/main/java/org/cloudfoundry/promregator/rewrite/MergableMetricFamilySamples.java
```diff
@@ -63,7 +63,7 @@ public void merge(HashMap<String,MetricFamilySamples> others) {
 		
 		for (Iterator<MetricFamilySamples> iterator = coll.iterator(); iterator.hasNext();) {
 			MetricFamilySamples mfs = iterator.next();
-			if (mfs.type == Type.UNTYPED) {
+			if (mfs.type == Type.UNKNOWN) {
```
checketts, line 66: With the Prometheus SImpleCLient upgrade `UNTYPED` went away so I replaced it with `UNKNOWN`

src/test/java/org/cloudfoundry/promregator/config/validations/PreferredRouteRegexMustBeCompilableTest.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.cloudfoundry.promregator.lite.config.Target
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.Test
+
+internal class PreferredRouteRegexMustBeCompilableTest {
+
+    @Test
+    fun testValidDoesNotBreak() {
+        val subject = PreferredRouteRegexMustBeCompilable()
+        val target = Target(preferredRouteRegex = listOf("dummy"))
```
checketts, line 13: Migrated this test to Kotlin so the simplified named parameter constructor could be used.

src/main/java/org/cloudfoundry/promregator/lite/cfaccessor/SimpleCFAccessorImpl.kt
```diff
@@ -0,0 +1,511 @@
...
+    private val rateLimiter = RateLimiter(600, meterRegistry)
+    private var _isV3Enabled = false
+    val isV3Enabled: Boolean get() = _isV3Enabled
+
+    private val cloudFoundryClients = ConcurrentHashMap<String, ReactorCloudFoundryClient>()
+
+    init {
+        meterRegistry.gauge("promregator.cf.ratelimit.available", rateLimiter) { it.availableTokens.toDouble() }
+        meterRegistry.gauge("promregator.cf.ratelimit.queue", rateLimiter) { it.waitQueue.toDouble() }
+        meterRegistry.more().counter("promregator.cf.ratelimit.wait_seconds", Tags.empty(), rateLimiter) { (it.totalWaitTimeMs / 1000.0) }
```
checketts, line 67: Here the ratelimiter metrics are exposed

src/main/java/org/cloudfoundry/promregator/lite/cfaccessor/SimpleCFAccessorImpl.kt
```diff
@@ -0,0 +1,511 @@
...
+        createClients()
+    }
+
+    private suspend fun <T> rateLimitAndRecord(requestType: RequestType, block: suspend () -> T): T? {
```
checketts, line 71: This method gates the ratelimit and records metrics

src/main/java/org/cloudfoundry/promregator/lite/cfaccessor/SimpleCFAccessorImpl.kt
```diff
@@ -0,0 +1,511 @@
...
+        return rateLimiter.whenTokenAvailable {
+            val timer = Timer.start()
+            val result = try {
+                 block()
+            } catch (e: Exception) {
+                log.error(e) { "Error making request type: $requestType"  }
+                null
+            }
+            timer.stop(meterRegistry.timer("promregator.cffetch.single", "type", requestType.metricName))
```
checketts, line 80: I separated out the `single` page calls as a separate timer since these added together would be double counted with the overall cffetch calls.

src/main/java/org/cloudfoundry/promregator/lite/cfaccessor/SimpleCFAccessorImpl.kt
```diff
@@ -0,0 +1,511 @@
...
+            result
+        }
+    }
+
+    private fun createClients() {
+        cf.api.forEach { api -> reset(api.key) }
```
checketts, line 86: Support for creating multiple clients (for talking to multiple APIs)

src/main/java/org/cloudfoundry/promregator/lite/cfaccessor/SimpleCFAccessorImpl.kt
```diff
@@ -0,0 +1,511 @@
...
+    private fun proxyConfiguration(apiConfig: ApiConfig): ProxyConfiguration? {
+        val effectiveProxyHost = apiConfig.proxy?.host
+        val effectiveProxyPort = apiConfig.proxy?.port ?: 0
+
+        if (effectiveProxyHost != null && PATTERN_HTTP_BASED_PROTOCOL_PREFIX.matcher(effectiveProxyHost).find()) {
+            throw ConfigurationException("Configuring of cf.proxyHost or cf.proxy.host configuration parameter must not contain an http(s)://-like prefix; specify the hostname only instead")
+        }
+        return if (effectiveProxyHost != null && effectiveProxyPort != 0) {
+            var proxyIP: String? = null
+            // TODO add back InetAddressUtils
```
checketts, line 141: I removed the InetAddressUtils since it had been the only usage of the apache lib in the lite project. I'll add it back.

src/main/java/org/cloudfoundry/promregator/lite/cfaccessor/SimpleCFAccessorImpl.kt
```diff
@@ -0,0 +1,511 @@
...
+//        val proxyConfiguration = proxyConfiguration()
+//        val connectionContext = connectionContext(proxyConfiguration)
+//        val tokenProvider = tokenProvider()
+//        cloudFoundryClient = cloudFoundryClient(connectionContext, tokenProvider)
+//    } catch (e: ConfigurationException) {
+//        log.error("Restarting Cloud Foundry Client failed due to Configuration Exception raised", e)
+//    }
+
+    suspend fun retrieveAllOrgIds(api: String): List<OrganizationResourceV2> {
+        return cfQueryV2<OrganizationResourceV2, ListOrganizationsResponseV2, ListOrganizationsRequestV2>(api, ORG) {
```
checketts, line 215: This is the new paginating mini-DSL. Just need to define a `request` and the `query` and it returns the `List<Resource>`.

src/main/java/org/cloudfoundry/promregator/lite/cfaccessor/SimpleCFAccessorImpl.kt
```diff
@@ -0,0 +1,511 @@
...
+            }
+            query { applicationsV2().list(it) }
+        }
+    }
+
+    suspend fun retrieveSpaceSummary(api: String, spaceId: String): GetSpaceSummaryResponse? {
+        // Note that GetSpaceSummaryRequest is not paginated
+        val request = GetSpaceSummaryRequest.builder().spaceId(spaceId).build()
+        return try {
+            cloudFoundryClients[api]?.spaces()?.getSummary(request)?.awaitSingle()
```
checketts, line 266: Note the `awaitSingle` call that converts the reactive call into a suspending coroutines call. Also error handling is managed by simple try/catch constructs.

src/main/java/org/cloudfoundry/promregator/lite/cfaccessor/SimpleCFAccessorImpl.kt
```diff
@@ -0,0 +1,511 @@
...
+    }
+
+    suspend fun retrieveAllDomains(api: String, orgId: String): org.cloudfoundry.client.v2.organizations.ListOrganizationDomainsResponse? {
+        val request = org.cloudfoundry.client.v2.organizations.ListOrganizationDomainsRequest.builder().organizationId(orgId).build()
+        return cloudFoundryClients[api]?.organizations()?.listDomains(request)?.awaitSingle()
+    }
+
+
+
+//    fun retrieveOrgIdV3(api: String, orgName: String?): Mono<ListOrganizationsResponse> {
```
checketts, line 280: I commented out the unused CF api calls. But will revisit them to setup and all v3 set of calls.

src/main/java/org/cloudfoundry/promregator/lite/cfaccessor/SimpleCFAccessorImpl.kt
```diff
@@ -0,0 +1,511 @@
...
+
+        fun request(block: (Int) -> S) {
+            requestTemplate = block
+        }
+
+        fun query(block: ReactorCloudFoundryClient.(S) -> Mono<T>) {
+            queryTemplate = block
+        }
+
+        suspend fun runQueries() : List<R> {
```
checketts, line 413: This method is where the brains of the query is.

It makes the `first` call to get the total number of pages (wrapping it in a ratelimit and timer) then fetches all the remaining pages in parallel. (See line 425)

src/main/java/org/cloudfoundry/promregator/lite/cfaccessor/SimpleCFAccessorImpl.kt
```diff
@@ -0,0 +1,511 @@
...
```
checketts, line 453: I had to create a v3 version of the same logic since it uses a different pagination class

checketts:
@eaglerainbow I'm eager to hear your thoughts. This is probably the fifth time I've tried to pull all my changes together to PR back upstream. The problem was how to introduce the new mechanisms without breaking or fighting the old ones.

Now that the logic is in a `lite` package, even though it duplicates functionality, it can evolve together and not keep getting stuck behind massive merge conflicts.

src/main/java/org/cloudfoundry/promregator/lite/discovery/SimpleTargetResolver.kt
checketts: When v3 is available, this path now uses all v3 endpoints! I need to fix the logic that calculates internal routes though.

eaglerainbow:
@checketts Wanted to say that I have seen your PR and started also reading through it. Given that there is also the other PR (where I am implicitly also a little side-effected on) and finding the time to work on these topics here is a little tricky for me right now (due to yet other activities), I have not come far yet enough yet to release my comments (I am somewhere at file 18 of 79 or so). Will try to stay on task...

checketts:
Thanks @eaglerainbow .  I completely understand.

checketts:
This PR has been rebased against master and is the code I'm using to monitor my clusters for the last few months.

I love how much easier everything is to setup now with the `http_sd_config` for service discovery and the annotations for target config. 

```yml
  - job_name: "promregator"
    scheme: https
    http_sd_configs:
      - url: https://<promregatorDomain>/v2/discovery
        
    relabel_configs:
     - source_labels: [__meta_promregator_target_applicationName,__meta_promregator_target_instanceNumber]
       target_label: instance
       separator: "-"
     - source_labels: [__meta_promregator_target_orgName]
       target_label: org_name
     - source_labels: [__meta_promregator_target_spaceName]
       target_label: space_name
       target_label: application
     - source_labels: [__meta_promregator_target_api]
       target_label: api
```

eaglerainbow:
> Are you OK with that approach?

Saw your comment, will come back to you as soon as possible.

eaglerainbow:
> Are you OK with that approach?

In general, approach is fine with me. 
So far the problem with this PR is that I am not able to digest the masses. I am currently thinking about it, what would be appropriate measures to make it more digestible (both volume and brain complexity)... Perhaps something like this?:

1. One PR introducing necessary infrastructure components for kotlin.
2. Transfer one rather simple thing over to kotlin (perhaps something like `AuthenticatorService` or `AuthenticatorEnricher`).
3. ... and then going step-by-step with next things

what do you think?

checketts:
Absolutely. I'm happy reset it and do smaller chunks. Now that is have the final state and have verified how it is simplified I can do it easily

eaglerainbow:
Great, but don't patch yourself too much forward - let's get to it better PR-by-PR

checketts:
That's the thing. I had been running on complete fork with little overlap, then I've been using this branch where at least I can align with current master. So doing small PRs to introduce this approach will be fine

PR #229, opened 2022-12-27
Lite kotlin

checketts:
This is a branch I created (and started using 1 year ago). It creates a 'lite' package that contains the core Promregator code and supports v3 for all endpoints.

It uses Kotlin coroutines to 'unwrap' the reactive CF Client and solves any leaks around it.


checketts:
No worries about merging this. Just wanted you to know it exists @eaglerainbow . Happy to rebase it against master if you like.

checketts:
I rebased this against master. And to clarify, the main app code is the same, the 'lite' package was originally the duplicated other app code then refactored and simplified. So both the old way and new way exist in the project.

The discovery v2 path triggers the new code, while discovery v1 keeps the old code. When we feel confident, then then we could make v1 call the new code and retire any un-needed code.

This also simplifies the migration path since we could warn on old config (or fail fast) directing people to the new config before removing it.

The only intermixing of old and new code is app config, since I wanted both paths to be able to leverage the same config.

